PR #1, opened 2017-05-03
Use APCU instead of classmap for composer

Pierstoval:
As you sweat to install APCU in the docker image, I think it might be a better idea to rely on apcu for autoload rather than classmap, because we can't use both, and apcu might be faster because memory.

What do you think?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,7 +61,7 @@ COPY src src/
 COPY var var/
 COPY web web/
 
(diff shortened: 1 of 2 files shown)


dunglas:
IIRC, classmap is faster than APCu because of the opcache. But I'm not sure :)

Pierstoval:
No, you're right. I made this PR before seeing that APCu class loader is useless with PHP7 because opcache works very well 😆

PR #2, opened 2017-06-12
Upgrade the Docker setup and dependencies

dunglas:


docker/nginx/conf.d/default.conf
```diff
@@ -1,12 +1,13 @@
 server {
-    root /srv/api-platform/web;
+    root /srv/app/web;
+    resolver 127.0.0.11;
```
jenkoian: Presumably a typo, but I had to change this to `127.0.0.1` to work. 👍

dunglas: It's not a typo: https://docs.docker.com/engine/userguide/networking/#embedded-dns-server

It should work as expected (and 127.0.0.1 shouldn't). Which version of Docker are you using?

jenkoian: `Docker version 17.03.1-ce, build c6d412e`

Removed all images and containers and trying again, will report back with any errors 👍

jenkoian: Yeah no dice, it just times out.

jenkoian: I have had friends try this out and it seems it may be an issue on Mac Sierra? Versions before this (e.g. El Capitan) it seems to work fine. Sorry, I can't offer more details at this time, but just FYI.

PR #3, opened 2017-06-17
Make specific access on dev environment optional

mickaelandrieu:
Why? Because as you now have only one front controller, it is *always* deployed in production :)

By default, this part should be now disabled and let advanced users write every custom action they need to do instead of force every Symfony developer to remove this damn piece of code every time on every project.

And don't tell me it's insecure: it's really hard to make ``APP_DEBUG`` env variable in production by mistake 👍


dunglas:
It is not straightforward for newcomers. You cannot start coding without Uncommenting those lines...
Usually you will not use this Docker compose file to deploy in production. APP_DEBUG will not be set until you do it explicitly, this code will not be executed.

mickaelandrieu:
> . You cannot start coding without Uncommenting those lines...

Why? It's quite the opposite! Most of the time, even if you have exported `APP_DEBUG` you still can't access your dev environment :/

dunglas:
It works on Linux and Mac but it seems to not work as expected on Windows.

dunglas:
Can you just update this PR to remove this block totally as I've done in the framework bundle's recipe?

Thanks Mickaël!

mickaelandrieu:
Yes Kevin, I'll do it probably tonight :)

Mickaël

mickaelandrieu:
Done (sorry for the delay) and thank you to have done it at Symfony's level :)

dunglas:
In fact, you can just remove the `web` directory now that this patch is in Flex!

mickaelandrieu:
Yeah, I wanted to ask you why we keep the web folder :D I do it now sir :)

dunglas:
> I don't know well Flex, will it create a new web folder for us?

Yes it will. No problem for this line.

dunglas:
Thanks @mickaelandrieu!

PR #5, opened 2017-07-06
Just type "docker-compose up" to install a Flex project (major refactoring)

dunglas:
This is a major refactoring of this repository. It allows to start hacking on a new Flex project by just downloading this repository and running `docker-compose up`.

There is no need to have PHP installed, the Docker setup will download and install Flex and create a new project during the first run. Project's files will be available both in the container and in the host (using a volume).

All PHP extensions (like APCu) and optimizations (like performance-related `php.ini` settings) are already handled.


docker/app/install-composer.sh
```diff
@@ -4,7 +4,7 @@
 # Origin: https://github.com/composer/composer/blob/master/doc/faqs/how-to-install-composer-programmatically.md
 # Licence: MIT
 
-EXPECTED_SIGNATURE=$(wget -q -O - https://composer.github.io/installer.sig)
```
teohhanhui: I must repeat this again. I don't see the point of us deviating from the official sample script here. The `.fetch-deps` do not end up in any of the image layers, so it has zero impact except for a very slightly longer build (does not make any practical difference).

teohhanhui: I thought we have already been through this before... *laughs*

dunglas: It reduces the complexity of the build process and the maintenance. As pointed out by Jordi, there is no point using the old method when using PHP 7+. And we're using PHP 7+.

teohhanhui: > It reduces the complexity of the build process and the maintenance.

Not really. `wget` is already provided by BusyBox. Installing `openssl` is required for HTTPS support. And that's it. I don't see where the complexity lies.

> As pointed out by Jordi, there is no point using the old method when using PHP 7+. And we're using PHP 7+.

Agreed 100%. But we don't really gain anything by changing it. It only creates another maintenance issue, if anything. (Instead of just copy-and-paste from the official sample script, now we have our own modifications. :laughing:)

PR #6, opened 2017-10-19
HTTP/2 and HTTPS support

dunglas:
Just open `https://localhost`.

* The HTTP/2 reverse proxy **is not secure** and is not suited for production usage. Prefer using CloudFlare or the reverse proxy provided by your cloud platform in production.
* Apache is used instead of Nginx as reverse proxy because Nginx doesn't support HTTP/2 Push yet bot Apache does (and this feature is present in Symfony: WebLink component)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,27 +37,25 @@ RUN set -xe \
 # https://getcomposer.org/doc/03-cli.md#composer-allow-superuser
 ENV COMPOSER_ALLOW_SUPERUSER 1
 
-RUN composer global require "hirak/prestissimo:^0.3" --prefer-dist --no-progress --no-suggest --optimize-autoloader --classmap-authoritative \
-	&& composer clear-cache
+# Use prestissimo to speed up builds
+RUN composer global require "hirak/prestissimo:^0.3" --prefer-dist --no-progress --no-suggest --optimize-autoloader --classmap-authoritative
+
+COPY docker/app/docker-entrypoint.sh /usr/local/bin/docker-app-entrypoint
+RUN chmod +x /usr/local/bin/docker-app-entrypoint
+
+# Download the Symfony skeleton and leverage Docker cache layers
+ENV SKELETON_COMPOSER_JSON https://raw.githubusercontent.com/symfony/skeleton/v3.3.4/composer.json
 
 WORKDIR /srv/app
+RUN php -r "copy('$SKELETON_COMPOSER_JSON', 'composer.json');" \
+    && composer install --prefer-dist --no-dev --no-progress --no-suggest --no-autoloader --no-scripts --no-plugins --no-interaction
 
 COPY . .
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -37,27 +37,25 @@ RUN set -xe \
...
+RUN composer global require "hirak/prestissimo:^0.3" --prefer-dist --no-progress --no-suggest --optimize-autoloader --classmap-authoritative
+
+COPY docker/app/docker-entrypoint.sh /usr/local/bin/docker-app-entrypoint
+RUN chmod +x /usr/local/bin/docker-app-entrypoint
+
+# Download the Symfony skeleton and leverage Docker cache layers
+ENV SKELETON_COMPOSER_JSON https://raw.githubusercontent.com/symfony/skeleton/v3.3.4/composer.json
 
 WORKDIR /srv/app
+RUN php -r "copy('$SKELETON_COMPOSER_JSON', 'composer.json');" \
```
greg0ire: I think you could use `ADD ` here

dunglas: No because the file is not included in the repo. But I’ll switch to composer create-project in another PR.

greg0ire: ? `ADD` works with remote urls...

dunglas: I wasn't aware of that! great

greg0ire: Yeah I was pretty happy too when I realised I didn't actually need curl or wget :)

B-Galati: Yes but from what I remember ADD cannot be cached as a layer by docker so it's quite a pain when you are rebuilding your image a lot. Perhaps it is not the case anymore.

dunglas: I think I've found a better way (no more `copy` nor `ADD`).

Dockerfile.h2-proxy
```diff
@@ -0,0 +1,25 @@
...
+        -key /usr/local/apache2/conf/server.key \
+        -out /usr/local/apache2/conf/server.csr \
+    && openssl x509 \
+        -req -sha256 \
+        -days 365 \
+        -in /usr/local/apache2/conf/server.csr \
+        -signkey /usr/local/apache2/conf/server.key \
+        -out /usr/local/apache2/conf/server.crt
+
+COPY ./docker/httpd/httpd.conf /usr/local/apache2/conf/httpd.conf
```
joelwurtz: You could maybe use here a multi stage build, first stage will create the certificate, second step will only copy them to apache so no need to have openssl as a dependency (unless it's required for apache to do https ?)

greg0ire: Doing so would allow you do avoid all these `&&` because you wouldn't care about having many layers anymore. :+1:

dunglas: Excellent idea @joelwurtz. It's done.

PR #11, opened 2017-10-26
Support building using specific symfony version

kejwmen:
Hi,
I've seen a question about installing specific symfony version on symfony slack and thought it could be supported here.

Not sure about `VERSION` default value in compose, but we'll probably figure it out.

What do you think?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,8 +39,11 @@ RUN composer global require "hirak/prestissimo:^0.3" --prefer-dist --no-progress
 ARG STABILITY=stable
 ENV STABILITY ${STABILITY}
 
+# Allow to select skeleton version
+ARG VERSION=""
+
 # Download the Symfony skeleton and leverage Docker cache layers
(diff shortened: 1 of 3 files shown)


dunglas:
`VERSION` must also be used here: https://github.com/dunglas/symfony-docker/blob/master/docker/app/docker-entrypoint.sh#L12

/cc @reypm

dunglas:
ping @kejwmen, do you want me to finish this one?

reypm:
@dunglas I am not the owner of this but yeah it would be great if you finish this and add it to the stack 💯

kejwmen:
Sorry, I forgot about this PR and thought there was a similiar one, already merged.

I will finish it later today.

kejwmen:
Done, sorry for delay ;)

kejwmen:
should we add a README to document this behavior?

dunglas:
Yes it would be awesome to add a README.

dunglas:
Thanks!

PR #18, opened 2018-06-14
Fix the build of app with auto-script like recipe framework-bundle

maxhelias:
Fix the build after installing the symfony skeleton caused by the composer auto-scripts that can not access the configuration

Example on my computer : https://ibb.co/dG70Ud


maxhelias:
Or rather that this PR, we can do with the '--no-scripts' option.

What do you prefer?

dunglas:
Can you switch to `--no-scripts` instead please? It will be better for prod images.

maxhelias:
Yes it's done. And i'm agree for prod image, i thought about it after

dunglas:
Thanks!

PR #21, opened 2018-06-21
Reverse proxy with nginx

maxhelias:
Replace the reverse proxy with nginx


Dockerfile.h2-proxy
```diff
@@ -10,8 +10,8 @@ RUN openssl req -new -passout pass:NotSecure -key server.key -out server.csr \
...
 
-FROM httpd:2.4-alpine
+FROM nginx:1.15-alpine
 
-COPY --from=0 server.key /usr/local/apache2/conf/server.key
-COPY --from=0 server.crt /usr/local/apache2/conf/server.crt
-COPY ./docker/httpd/httpd.conf /usr/local/apache2/conf/httpd.conf
+RUN mkdir -p /etc/nginx/ssl/
+COPY --from=0 server.key server.crt /etc/nginx/ssl/
+COPY ./docker/h2-proxy/default.conf /etc/nginx/conf.d/default.conf
```
dunglas: Missing blank line at the end

docker/h2-proxy/default.conf
```diff
@@ -0,0 +1,12 @@
+server {
+    listen 443 ssl http2;
+    listen [::]:443 ssl http2;
+
+    ssl_certificate /etc/nginx/ssl/server.crt;
+    ssl_certificate_key /etc/nginx/ssl/server.key;
+    ssl_session_cache  builtin:1000  shared:SSL:10m;
```
dunglas: I wouldn't change the default value (as this container is only intended for dev, we can keep the config minimal).

Dockerfile.h2-proxy
ajardin: Don't forget the newline at the end of file. 😉

dunglas:
@teohhanhui, would you mean reviewing this one?

docker-compose.yaml
```diff
@@ -35,6 +35,6 @@ services:
       context: .
       dockerfile: ./Dockerfile.h2-proxy
     volumes:
-      - ./docker/httpd/httpd.conf:/usr/local/apache2/conf/httpd.conf:ro
+      - ./docker/h2-proxy/default.conf:/etc/nginx/conf.d/default.conf:ro
```
ajardin: As the file seems to be copied during the image building, is it really needed?

dunglas: Yes. Copying is for prod, the volume is for dev (you change the file, reboot the service and it's taken into account without having to rebuild the image)

ajardin: That's my first guess. 😄 
Maybe we can add the same notice as for the nginx service (`# Comment out the next line in production`).

maxhelias:
@dunglas it's good 😃

B-Galati:
What about doing the same thing in the Dockerfile.nginx ?
Would it be wrong or impossible ?

maxhelias:
@B-Galati It is not secure for use in production because it is a self-generated certificate.
We use the h2-proxy service only for development

B-Galati:
@maxhelias Thanks for your answer.

Considering a development environment only, does it make sense to keep both `nginx` and `h2-proxy` service while `nginx` could do both ?

maxhelias:
@B-Galati Yes, if you use it only in development you can configure your h2 directly in the nginx service.

maxhelias:
With the current configuration, you can use it in both cases.
Just pay attention to the docker-compose's comments about the production before pushing the images

dunglas:
Thanks @maxhelias!

PR #24, opened 2018-06-25
Improve OPcache settings

ajardin:
`opcache.memory_consumption` is present on the [Symfony documentation](http://symfony.com/doc/current/performance.html#configure-opcache-for-maximum-performance) while `opcache.interned_strings_buffer` should be upgraded also as the default value is quite low.

### DIFF
--- a/docker/app/php.ini
+++ b/docker/app/php.ini
@@ -4,6 +4,8 @@ session.auto_start = Off
 short_open_tag = Off
 
 # http://symfony.com/doc/current/performance.html
+opcache.interned_strings_buffer = 16
 opcache.max_accelerated_files = 20000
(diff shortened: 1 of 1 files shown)


bigfoot90:
Also disable file timestamps if in production environment
http://symfony.com/doc/current/performance.html#don-t-check-php-files-timestamps

ajardin:
@bigfoot90 I was also tempted to go further, but adding that parameter implies to manually update the PHP configuration on development environments.
The solution could be to maintain two configuration files in parallel.

dunglas:
Thanks @ajardin! We'll be glad to merge the same changes in API Platform https://github.com/api-platform/api-platform/blob/master/api/docker/php/php.ini

PR #25, opened 2018-06-25
Pass stability from env variable to dockerfile

kejwmen:
Usage of this feature should be consistent with selecting specific version.


dunglas:
Thanks!

PR #26, opened 2018-06-26
Gather all Dockerfile in one

maxhelias:
Use a single Dockerfile


ajardin:
That's only my personal point of view, but I'm generally against this kind of writing since the final result is less readable. 😕 

@dunglas Is there any specific benefit (workflow, performance, etc.) for merging everything into a single file? I saw you did the same change for API Platform a couple of days ago.

jorge07:
Docker is about process isolation. Using docker as a Vagrant box is a mistake.

dunglas:
@jorge07 process are still isolated. It's one file that creates several containers.
@ajardin It's more in the spirit of Flex (less files, minimal setup), only the minimal things required to get started. Also it allows to be in sync with `api-platform/api-platform`: https://github.com/api-platform/api-platform/pull/666

jorge07:
@dunglas I didn't saw you're using `target` so yes, my fault. 
Still having my concerns about having a single Dockerfile for all the stack. Where is single responsibility then?
Any blog post or related explaining pros-cons?

johnutzm:
Currently Docker does not support skipping stages (docker/cli#1134). If you want to build the last target in the Dockerfile everything will be built.

dunglas:
Thanks @maxhelias!

PR #30, opened 2018-08-23
Fix the STABILITY env

maxhelias:
Fix the error at "docker-compose up" by making STABILITY optional


dunglas:
Thanks @maxhelias!

PR #33, opened 2018-08-25
Fix docker-compose after all dockerfile in one

maxhelias:
I forgot to give the target in docker-compose.yaml file after my tests


dunglas:
Thanks Max!

PR #37, opened 2018-10-16
Updated readme to include a note about linux users

vonalbert:
Added instruction to fix filesystem permissions on linux, that right after the project creation do not allow the user in the host system to write files.

Related to issue #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
     with Xdebug v2.6.0, Copyright (c) 2002-2018, by Derick Rethans
 ```
 
-### Editing permissions on Linux
+### Editing Permissions on Linux
 
 If you work on linux and cannot edit some of the project files right after the first installation, you can run `docker-compose run --rm app chown -R $(id -u):$(id -g) .` to set yourself as owner of the project files that were created by the docker container.
 
--- a/README.md
+++ b/README.md
@@ -8,6 +8,10 @@ A [Docker](https://www.docker.com/)-based installer and runtime for the [Symfony
 2. Open `https://localhost` in your favorite web browser and [accept the auto-generated TLS certificate](https://stackoverflow.com/a/15076602/1352334)
 3. **Enjoy!**
 
+### Note for linux users
+
+If you work on linux and cannot edit some of the project files right after the first installation, you can run `docker-compose run --rm app chown -R $(id -u):$(id -g) .` to set yourselft as owner of the project files that were created by the docker container.
+
 ## Selecting a Specific Symfony Version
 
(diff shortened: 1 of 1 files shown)


vonalbert:
Since we're using docker compose i proposed a docker-compose version that uses the app container to execute the command. This results in a  cleaner and more readable command.

`docker-compose run --rm app chown -R $(id -u):$(id -g) . `
compared to
`docker run --rm -v $(pwd):/app -w /app busybox chown -R $(id -u):$(id -g) .`

README.md
```diff
@@ -8,6 +8,10 @@ A [Docker](https://www.docker.com/)-based installer and runtime for the [Symfony
 2. Open `https://localhost` in your favorite web browser and [accept the auto-generated TLS certificate](https://stackoverflow.com/a/15076602/1352334)
 3. **Enjoy!**
 
+### Note for linux users
```
dunglas: May I suggest:

    ### Editing `vendor/` and `var/` directories on Linux

Can you also add this section just before the "Credits" one?

README.md
```diff
@@ -8,6 +8,10 @@ A [Docker](https://www.docker.com/)-based installer and runtime for the [Symfony
...
+
+If you work on linux and cannot edit some of the project files right after the first installation, you can run `docker-compose run --rm app chown -R $(id -u):$(id -g) .` to set yourselft as owner of the project files that were created by the docker container.
```
dunglas: yourself

README.md
vonalbert: Well the problem here is that is not only the vendor and var directory is not writable, but even the src and config because the very first time you execute `docker-compose up` these files doesn't exist, thus are created when containers are started (by the entrypoint).

So to reflect this the title could become
```
### Editing source files on Linux
```

Ok so I'll add the section as the last h3 of the Debugging section, right after troubleshooting.

README.md
vonalbert: Good catch

vonalbert:
The section title still need to be updated, but I wait for feedback before doing it

README.md
maxhelias: The command edits all the files of the container ?
If so, in which case this would agree more :
```
### Editing permissions on Linux
```
WDYT ?

vonalbert: Yes, this command sets your user as owner for all the files through the container.
This is useful only for dev - when you use mounted volumes - because you change project files in your host machine as well;
in production (when you shouldn't use volumes) it's not really useful because you can handle permission issues directly in dockerfile, like you already do [here](https://github.com/dunglas/symfony-docker/blob/master/Dockerfile#L87)

However `Editing permissions on Linux` as a section title is fine for me.

dunglas: Go for `Editing Permissions on Linux` (in title case please). Thanks!

dunglas:
Thanks @vonalbert

PR #40, opened 2019-01-06
Use php.ini depending on env

maxhelias:
It's like api-platform/api-platform#857 with the improvement of api-platform/api-platform#892


dunglas:
Thanks!

PR #41, opened 2019-01-06
Adding Travis integration

maxhelias:
We can not use the docker addons because that provide a docker-compose version (1.17.1) with a bug on the variables substitution.
So I installed the docker version corrected by my self


dunglas:
Awesome! Thank you very much Max!

PR #42, opened 2019-01-17
fix service name in docker-compose command

FabienSalles:
Replace the service name in docker-compose command with the correct one


dunglas:
Good catch, thanks!

PR #44, opened 2019-01-25
Simplify docker-compose command with override file

gmsantos:
Docker compose automatically merge both `docker-compose.yaml` and `docker-compose.override.yaml` files if `-f`` is not provided.

So I simplified the command and make a quick note inspired from docker compose documentation (https://docs.docker.com/compose/reference/overview/#specifying-multiple-compose-files)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,8 +79,10 @@ services:
 Then run:
 
 ````bash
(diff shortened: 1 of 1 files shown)


maxhelias:
Thanks for the contribution and the information 😃

PR #47, opened 2019-04-09
Add information to run behind a corporate proxy

patrick-mota:
Even if it's not a bug and it's more related about general proxy knowledge (in and outside Docker), I think it's always good to add some information if someone (like me) wanted to run it behind a corporate proxy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,22 @@ For instance, use the following command to use the `master` branch of Symfony:
 STABILITY=dev docker-compose up --build
 ```
 
+## Build and running behind a proxy
+
+1. Uncomment in the Dockerfile the following line
+```bash
+RUN pear config-set http_proxy yourproxy
+```
+2. Replace "yourproxy" by your proxy
+3. Build
+```bash
+docker-compose build --build-arg http_proxy="yourproxy" --build-arg https_proxy="yourproxy"
+```
+4. Up
+```bash
+docker-compose up
+```
+
 ## Debugging
 
 The default Docker stack is shipped without a Xdebug stage.
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,6 +38,10 @@ RUN apk add --no-cache --virtual .persistent-deps \
 		zlib
 
 ENV APCU_VERSION 5.1.12
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -25,6 +25,22 @@ For instance, use the following command to use the `master` branch of Symfony:
...
+## Build and running behind a proxy
+
+1. Uncomment in the Dockerfile the following line
+```bash
+RUN pear config-set http_proxy yourproxy
+```
+2. Replace "yourproxy" by your proxy
+3. Build
+```bash
+docker-compose build --build-arg http_proxy="yourproxy" --build-arg https_proxy="yourproxy"
```
maxhelias: I have never used pear through a proxy and I don't know this part very well but can we expose http_proxy & https_proxy in the Dockerfile and use the ENV variable like this : `PROXY="http://yourproxy.com/" docker-composer build` ?
This would delete step 2

README.md
```diff
@@ -25,6 +25,22 @@ For instance, use the following command to use the `master` branch of Symfony:
...
+```
+4. Up
+```bash
+docker-compose up
```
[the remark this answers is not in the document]
patrick-mota: Usually I use -d too but since in the doc. it was already using docker-compose up without the -d I thought it will be better to keep the same behaviour.

README.md
patrick-mota: @maxhelias for the docker-compose I'm not sure you can build this way, isn't env var but a build-arg.
If I understand using PROXY="" before docker-compose will set a ENV proxy to docker-compose which can be used when building a Dockerfile but I don't know if it can be used this way with docker-compose.

Anyway, we can use the env and inject it into the Dockerfile when we use pear.

maxhelias: For `PROXY`, i mean like `SYMFONY_VERSION` https://github.com/dunglas/symfony-docker/blob/d5cc26520a156731110e5c3cf486d2aa056ff5e9/Dockerfile#L76

https://github.com/dunglas/symfony-docker/blob/d5cc26520a156731110e5c3cf486d2aa056ff5e9/Dockerfile#L79

PR #50, opened 2019-05-27
Update Dockerfile

maxgio92:
- Add libzip-dev system package
- Configure zip PHP extension with libzip


dunglas:
Why is this extension needed by default?

maxgio92:
@dunglas , sorry, I wrongly closed the PR.
The problem occurs with PHP v7.3.5, not with PHP v7.2, on 'docker-php-ext-install zip':

`checking for libzip... not found
configure: error: Please reinstall the libzip distribution`

PR #51, opened 2019-06-27
Upgrade Dockerfile

dunglas:
Syncrhonize the Dockerfile with the last version in API Platform (/cc @teohhanhui)


Dockerfile
```diff
@@ -84,14 +79,53 @@ ENV STABILITY ${STABILITY:-stable}
...
+# Use this self-generated certificate only in dev, IT IS NOT SECURE!
+RUN openssl genrsa -des3 -passout pass:NotSecure -out server.pass.key 2048
+RUN openssl rsa -passin pass:NotSecure -in server.pass.key -out server.key
+RUN rm server.pass.key
+RUN openssl req -new -passout pass:NotSecure -key server.key -out server.csr \
+	-subj '/C=SS/ST=SS/L=Gotham City/O=Symfony/CN=localhost'
+RUN openssl x509 -req -sha256 -days 365 -in server.csr -signkey server.key -out server.crt
+
+### "h2-proxy" stage
+FROM nginx:${NGINX_VERSION}-alpine AS symfony_docker_h2-proxy
```
teohhanhui: Do we really need this?

README.md
```diff
@@ -100,7 +100,7 @@ Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
 
 ### Editing Permissions on Linux
 
-If you work on linux and cannot edit some of the project files right after the first installation, you can run `docker-compose run --rm app chown -R $(id -u):$(id -g) .` to set yourself as owner of the project files that were created by the docker container.
+If you work on linux and cannot edit some of the project files right after the first installation, you can run `docker-compose run --rm php chown -R $(id -u):$(id -g) .` to set yourself as owner of the project files that were created by the docker container.
```
teohhanhui: I don't think this would work correctly. The uid and gid in the container is that of `root`.

docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
-version: '3.4'
+version: "3.4"
```
teohhanhui: This file should be renamed to `docker-compose.yml` to reduce friction.

docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
...
 
 services:
-  app:
+  php:
```
teohhanhui: The part about volumes for Linux should be synced from api-platform/api-platform too.

docker/php/docker-entrypoint.sh
```diff
@@ -23,12 +23,12 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'bin/console' ]; then
         cp -Rp tmp/. .
         rm -Rf tmp/
     elif [ "$APP_ENV" != 'prod' ]; then
-        # Always try to reinstall deps when not in prod
         composer install --prefer-dist --no-progress --no-suggest --no-interaction
     fi
 
-	# Permissions hack because setfacl does not work on Mac and Windows
-	chown -R www-data var
+	mkdir -p var/cache var/log
```
teohhanhui: This should be moved up before the previous `if` block.

docker/php/docker-entrypoint.sh
```diff
@@ -6,12 +6,12 @@ if [ "${1#-}" != "$1" ]; then
...
-    fi
-    ln -sf "$PHP_INI_RECOMMENDED" "$PHP_INI_DIR/php.ini"
+if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
+	PHP_INI_RECOMMENDED="$PHP_INI_DIR/php.ini-production"
+	if [ "$APP_ENV" != 'prod' ]; then
+		PHP_INI_RECOMMENDED="$PHP_INI_DIR/php.ini-development"
+	fi
+	ln -sf "$PHP_INI_RECOMMENDED" "$PHP_INI_DIR/php.ini"
 
     # The first time volumes are mounted, the project needs to be recreated
```
teohhanhui: What is the reason for this? Let's see if we could figure this one out.

Dockerfile
```diff
@@ -84,14 +79,53 @@ ENV STABILITY ${STABILITY:-stable}
 ARG SYMFONY_VERSION=""
 
 # Download the Symfony skeleton and leverage Docker cache layers
-RUN composer create-project "symfony/skeleton ${SYMFONY_VERSION}" . --stability=$STABILITY --prefer-dist --no-dev --no-progress --no-scripts --no-plugins --no-interaction
+RUN composer create-project "symfony/skeleton ${SYMFONY_VERSION}" . --stability=$STABILITY --prefer-dist --no-dev --no-progress --no-scripts --no-plugins --no-interaction; \
```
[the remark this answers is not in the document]
teohhanhui: > I've tried with composer create-project symfony-skeleton:dev-master and it works.

Because there's `"minimum-stability": "dev"` in the unreleased versions of `symfony/skeleton`. But yeah, `minimum-stability` is dangerous and we should not encourage setting it to anything other than `stable` (the default). So we still should not use `--stability` here.

docker/php/docker-entrypoint.sh
dunglas: Because the files of the project itself aren't present on the host, and you cannot (AFAIK) create a volume sharing file from the inside of the container to the host, you have to do it the other way around.

dunglas: Maybe could we create an internal volume with `VOLUME` and then mount it with Docker Compose? I'm not sure if it could work, but it could be tried.

teohhanhui: Actually, we should not run `composer create-project` inside Dockerfile. It makes no sense. If it's to make sure the project is in sync with `symfony/skeleton`, it should be done externally.

dunglas: But doing it "externally" is by nature incompatible with the main objective of this project: you must be able to bootstrap a new Symfony project by just running `docker-compose up`, without having PHP, Composer, Bash or anything else installed on the host computer.
The goal is to provide a similar experience than with API Platform, or Laravel Homestead.

If we assume that the user has PHP and Composer installed, we could just ship the relevant Dockerfile and `docker-compose.yaml` files in the recipes.

teohhanhui: Sorry for the confusion. I meant the syncing must be done externally (maybe by a bot operated by the Symfony team). The project must indeed be self-contained. Currently that's not the case, because you're creating the project when building the Docker image.

dunglas: The current approach is definitely a hack.
Creating a bot that regenerating this skeleton every time recipe change (and when a new Symfony version is released) would be definitely cleaner (it was my 1st idea), however when I created this repository, I haven't managed a good way to detect such events in an easy way.
While thinking about it, I'm sure @fabpot already has some tools like that up and running for Flex and the existing skeletons. Fabien, do you think we could trigger such bot on an existing hook?

Dockerfile
```diff
@@ -1,80 +1,75 @@
...
-# Use this self-generated certificate only in dev, IT IS NOT SECURE!
-RUN openssl genrsa -des3 -passout pass:NotSecure -out server.pass.key 2048
-RUN openssl rsa -passin pass:NotSecure -in server.pass.key -out server.key
-RUN rm server.pass.key
-RUN openssl req -new -passout pass:NotSecure -key server.key -out server.csr \
-    -subj '/C=SS/ST=SS/L=Gotham City/O=Symfony/CN=localhost'
-RUN openssl x509 -req -sha256 -days 365 -in server.csr -signkey server.key -out server.crt
+# https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact
+ARG PHP_VERSION=7.3
+ARG NGINX_VERSION=1.15
```
maxhelias: You can also bump nginx to 1.17

README.md
```diff
@@ -55,7 +55,7 @@ configuration remains untouched.
 As example, an override could look like this:
 
 ```yaml
-version: '3.4'
+version: "3.4"
 
 services:
   app:
```
maxhelias: We also need to update the service name here

Dockerfile
```diff
@@ -1,80 +1,75 @@
...
+	jq \
+	;
+
+ARG APCU_VERSION=5.1.17
+RUN set -eux; \
+	apk add --no-cache --virtual .build-deps \
+	$PHPIZE_DEPS \
+	icu-dev \
+	libzip-dev \
+	postgresql-dev \
```
alanpoulain: Should be removed.

Dockerfile
```diff
@@ -84,14 +79,53 @@ ENV STABILITY ${STABILITY:-stable}
...
+CMD ["php-fpm"]
+
+
+# "nginx" stage
+# depends on the "php" stage above
+FROM nginx:${NGINX_VERSION}-alpine AS symfony_docker_nginx
+
+COPY docker/nginx/conf.d/default.conf /etc/nginx/conf.d/default.conf
+
+WORKDIR /srv/app
```
alanpoulain: Don't you want to change the workdir as well?

Dockerfile
```diff
@@ -1,80 +1,75 @@
...
-    && apk add --no-cache --virtual .api-phpexts-rundeps $runDeps \
-	&& apk del .build-deps
+	acl \
+	file \
+	gettext \
+	git \
+	jq \
+	;
+
+ARG APCU_VERSION=5.1.17
```
alanpoulain:
```suggestion
ARG APCU_VERSION=5.1.18
```

alanpoulain:
It would be great if this PR was merged!

README.md
maxhelias: As we already have this note, that should not block, we will try that works well after the merger of this PR.

docker-compose.yaml
maxhelias: I'm not sure, should we delete the lines?

maxhelias:
@dunglas, @teohhanhui, @alanpoulain, I updated the branch with all your feedback. Can you review again ?

dunglas:
lgtm

maxhelias:
Thanks !

PR #54, opened 2019-10-02
Speed up using the host composer cache

itsazzad:


docker-compose.yaml
```diff
@@ -9,6 +9,7 @@ services:
         SYMFONY_VERSION: ${SYMFONY_VERSION:-}
         STABILITY: ${STABILITY:-stable}
     volumes:
+      - ~/.composer/cache:/root/.composer/cache
```
dunglas: Will it work on Windows?

itsazzad: Not tested

dunglas: I fear it will not.

maxhelias:
This doesn't work on Windows, and there is currently no solution for compatibility between the different envs.

PR #56, opened 2019-12-20
Fix --with-libzip

maxhelias:
--with-libzip was removed with PHP 7.4 and I don't know why the build on the initial PR was green


maxhelias:
ah wrong branch 😅

Issue #59, opened 2020-05-08
Symfony profiler pointing to nginx/_profiler/


thePanz:
When using the application from `https://localhost` and enabling the Symfony profiler-pack, the web-toolbar is pointing to `nginx/_profiler/xxxxx`

The issue happens only when using the `https` protocol, under `http` the link works correctly and points to `localhost/_profiler/xxxx`

maxhelias:
@thePanz can you confirm that the #60 will correct the problem?

maxhelias:
Fixed by #60

PR #60, opened 2020-05-12
Add missing proxy options

maxhelias:
These options sould be fix the issue #59


thePanz:
Yes, tested and it works now! Thank you @maxhelias

PR #71, opened 2020-10-14
feat: add support for PHP preloading

dunglas:
See https://symfony.com/blog/new-in-symfony-5-1-configurable-php-preloading.


hacfi:
Hasn't the path to the `config/preload.php` changed in the meantime? See https://symfony.com/blog/php-preloading-and-symfony

PR #74, opened 2020-10-15
feat: add support for Doctrine ORM

dunglas:
Needs https://github.com/symfony/recipes/pull/113.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       start_period: 30s
     environment:
       # Run "composer require symfony/orm-pack" to install and configure Doctrine ORM
-      DATABASE_URL: mysql://symfony:ChangeMe@db:3306/app?serverVersion=mariadb-10.5.6
+      DATABASE_URL: ${DATABASE_URL:-mysql://symfony:ChangeMe@db:3306/app?serverVersion=mariadb-10.5.6}
       SYMFONY_VERSION:
 
   caddy:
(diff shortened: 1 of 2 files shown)


maxhelias (approved):

docker-compose.yml
```diff
@@ -15,6 +15,8 @@ services:
       retries: 3
       start_period: 30s
     environment:
+      # Run "composer require symfony/orm-pack" to install and configure Doctrine ORM
+      DATABASE_URL: mysql://symfony:ChangeMe@db:3306/app?serverVersion=mariadb-10.5.6
```
maxhelias, line 19: Now the variable can be in the override ? And maybe write it like `${DATABASE_URL:-mysql://symfony:ChangeMe@db:3306/app?serverVersion=mariadb-10.5.6}` ?

docker-compose.yml
```diff
@@ -15,6 +15,8 @@ services:
       retries: 3
       start_period: 30s
     environment:
+      # Run "composer require symfony/orm-pack" to install and configure Doctrine ORM
+      DATABASE_URL: postgresql://${POSTGRES_USER:-symfony}:${POSTGRES_PASSWORD:-ChangeMe}@db:5432/${POSTGRES_DB:-app}?serverVersion=${POSTGRES_VERSION:-13}
```
hacfi, line 19: Maybe add an additional commented out line with mysql/mariadb configuration? I see that question way too often on the Symfony Devs Slack channel.

dunglas: There is already a message explaining how to configure MySQL/Maria in the recipe's post install message.

For Docker, a comment will not be enough because the Dockerfile and `docket-compose.yml` must also be updated.

Issue #79, opened 2020-10-28
Trust root certificate locally


jessedobbelaere:
Great repository! 

### localhost
I was fiddling with a Symfony app and Caddyserver locally. One thing I haven't cracked is to have trusted SSL certificates. The link in the README to [accept the auto-generated TLS certificate](https://stackoverflow.com/a/15076602/1352334) seems outdated and does not work very well on my Mac with Chrome :( 

![image](https://user-images.githubusercontent.com/1352979/97448982-80354a00-1931-11eb-9fdc-6559ec171c35.png)

Any recent tricks or updates on how to make it work to self-sign the Caddy certificate? **Maybe I would need to somehow extract the Caddy root certificate out of the docker container and trust it on my mac...** Like `caddy trust` does...  I'd still like to let Caddy create the certificate, so I don't have to mount certificates created via e.g. https://github.com/FiloSottile/mkcert

### hostname
In addition, I'd like to use a hostname locally, e.g. `SERVER_NAME=mywebsite.test docker-compose up`. And I run dnsmasq to map *.test domains to my localhost. But the Caddy server will start reaching out to Letsencrypt (which won't work because the hostname is not externally available). 

```
caddy_1  | {"level":"error","ts":1603895023.9278789,"logger":"tls.obtain","msg":"will retry","error":"[mywebsite.test] Obtain: [mywebsite.test] creating new order: request to https://acme-staging-v02.api.letsencrypt.org/acme/new-order failed after 1 attempts: HTTP 400 urn:ietf:params:acme:error:rejectedIdentifier - Error creating new order :: Cannot issue for \"mywebsite.test\": Domain name does not end with a valid public suffix (TLD) (ca=https://acme-staging-v02.api.letsencrypt.org/directory)","attempt":3,"retrying_in":120,"elapsed":183.5167383,"max_duration":2592000}
```

Probably this can be fixed by adding `tls internal` to the Caddyfile. But then I still have the self-signed issue from above again.

maxhelias:
Hello @jessedobbelaere,

For the trust certificate, I have updated the documentation on this subject so that it now corresponds to the use of Caddy. Please see https://github.com/dunglas/symfony-docker/pull/85

For the hostname, the project is designed to be production, development and CI ready, so I'm afraid we don't put `tls internal`.

jessedobbelaere:
Late answer, but I just tested it out, works perfectly with the `docker cp` solution! Awesome 🙌  thank you!

PR #85, opened 2020-11-16
Update SSL trust issue

maxhelias:
Fix https://github.com/dunglas/symfony-docker/issues/79

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -6,6 +6,9 @@ If you work on linux and cannot edit some of the project files right after the f
 
 ## Fix Chrome/Brave SSL
 
-If you have a SSL trust issues, download the self-signed certificate and run :
+If you have a SSL trust issues, you can copy the self-signed certificate from caddy and add it to the trusted certificates :
 
-    $ sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain /path/to/you/certificate.cer
+    # MacOS
+    $ docker cp $(docker-compose ps -q caddy):/data/caddy/pki/authorities/local/root.crt /tmp/root.crt && sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain /tmp/root.crt
+    # Linux
+    $ docker cp $(docker-compose ps -q caddy):/data/caddy/pki/authorities/local/root.crt /usr/local/share/ca-certificates/root.crt && sudo update-ca-certificates


thePanz:
@maxhelias the mentioned command `sudo security add-trusted-cert ` is not recognized under Linux (do we need to install a separated tool for that?)

maxhelias:
Yes I miss that 😅 
For linux, I think it should be like that : 
```
sudo cp ./docker/cert/local/root.crt /usr/local/share/ca-certificates/foo.crt && sudo update-ca-certificates
```

Can you confirm, please ?

docker/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Used for copy the self-signed certificate from caddy and add it to the trusted certificates
+# see https://github.com/dunglas/symfony-docker/blob/master/docs/troubleshooting.md#fix-chromebrave-ssl
+cert/*
```
dunglas, line 3: Do we need to create a temporary directory? Can't we use a pipe or a temporary file?

B-Galati:
What about using something like mkcert ?

maxhelias:
@B-Galati, In fact Caddy already does the job but inside the container. The problem we have is that the certificate is not approved on the local machine.
Is it possible to copy it and trust it with mkcert?

B-Galati:
@maxhelias I am not sure, that may be possible by copying the Caddy certificate at right place.

In the end, I think that it would be simpler to use a local certificate provided by `mkcert` than using the one provided by Caddy.

It would require to add a new volume with the generated certificates and use them in Caddy config.

docker/.gitignore
maxhelias: It's now in tmp. Are these command good ?

maxhelias:
I re-open this PR in #93 .. I don't know why the conflict resolution was not updating here.

@B-Galati It could be a solution, feel free to open a issue or PR to have it in the pipeline. 
On the other hand, I don't see how we can do this while keeping the first idea of ​​the project which is to have an environment ready with simply "docker-compose up" (like api-platform distribution). But I would be happy if we can find an effective solution.

B-Galati:
> On the other hand, I don't see how we can do this while keeping the first idea of ​​the project which is to have an environment ready with simply "docker-compose up" (like api-platform distribution). But I would be happy if we can find an effective solution.

@maxhelias Agree :heavy_check_mark: 

If it can help I wrote a script that install mkcert in the current directory if not installed globally: https://gist.github.com/B-Galati/eb1f58de5235a9abe4277eaf946fb85a

thePanz:
@B-Galati have a look at #106 :)
I was able to get the mkcert certs working with Caddy, with no impact on the `production` environment :)

PR #87, opened 2020-11-17
Remove global Flex as it's not needed anymore with Composer 2

fabpot:
As the Docker image is now using Composer 2, this trick is not needed anymore.


dunglas (approved):

dunglas:
Thanks @fabpot!

PR #93, opened 2020-12-04
fix: update SSL trust issue

### LINKED ISSUE #79: Trust root certificate locally

maxhelias:
Fix #79

It's a re-open of #85.. I don't know why the conflict resolution was not updating on the origianl PR

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -6,11 +6,14 @@ If you work on linux and cannot edit some of the project files right after the f
 
 ## Fix Chrome/Brave SSL
 
-If you have a SSL trust issues, download the self-signed certificate and run :
+If you have a TLS trust issues, you can copy the self-signed certificate from Caddy and add it to the trusted certificates :
 
-    $ sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain /path/to/you/certificate.cer
+    # Mac
+    $ docker cp $(docker-compose ps -q caddy):/data/caddy/pki/authorities/local/root.crt /tmp/root.crt && sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain /tmp/root.crt
+    # Linux
+    $ docker cp $(docker-compose ps -q caddy):/data/caddy/pki/authorities/local/root.crt /usr/local/share/ca-certificates/root.crt && sudo update-ca-certificates
 
 ## HTTPs and Redirects
 
 If Symfony is generating an internal redirect for an `https://` url, but the resulting url is `http://`, you have to uncomment the `TRUSTED_PROXIES` setting in your `.env` file.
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -6,11 +6,14 @@ If you work on linux and cannot edit some of the project files right after the f
 
 ## Fix Chrome/Brave SSL
 
-If you have a SSL trust issues, download the self-signed certificate and run :
+If you have a SSL trust issues, you can copy the self-signed certificate from caddy and add it to the trusted certificates :
```
dunglas, line 9:
```suggestion
If you have a TLS trust issues, you can copy the self-signed certificate from Caddy and add it to the trusted certificates:
```

docs/troubleshooting.md
```diff
@@ -6,11 +6,14 @@ If you work on linux and cannot edit some of the project files right after the f
...
 
-    $ sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain /path/to/you/certificate.cer
+    # MacOS
```
dunglas, line 11:
```suggestion
    # Mac
```

dunglas (approved):

maxhelias:
Thanks @dunglas for you feedback :)

PR #103, opened 2021-01-08
Rename docker-compose*.yml docker-compose*.yaml

lyrixx:
In order to ease `bin/console make:docker:database` because it uses `.yaml` file


maxhelias (approved):

dunglas:
I would prefer to keep it as is. In their docs Docker Compose uses `.yml` and it's also the most used extension in various docs and tools.

DuboisS:
Yes, rename it directly in maker-bundle.

weaverryan:
We can (should) also make MakerBundle aware of both formats - have it look for both. That should be an easy, ideal solution I think.

dunglas:
Closing as this will handled directly in MakerBundle.

PR #105, opened 2021-01-13
fix: experimental_http3 moved under servers.protocol

thePanz:
fix `experimental_http3` deprecation (moved under `server.protocol` "namespace")


dunglas (approved):

dunglas:
Thanks @thePanz

PR #107, opened 2021-01-14
fix: properly check for DB environment

thePanz:
Make sure the `DATABASE_URL` is checked without comments :)


dunglas (approved):

dunglas:
Good idea. We should backport this in API Platform too.

PR #110, opened 2021-01-22
fix: do not check DB availability if code errors

thePanz:
Do not wait for the DB to be available if there are other errors blocking the access to it.

How to test it: 
- Create an application with a DB
- Add a PHP error in our files (controller, commands, ....)
- Restart the php container


dunglas:
Nice! I also updated this in API Platform to display a better error message in case of SQL error, could you try to merge both? https://github.com/api-platform/api-platform/blob/main/api/docker/php/docker-entrypoint.sh#L27

thePanz:
@dunglas updated :)

dunglas:
Thanks @thePanz!

PR #116, opened 2021-03-18
add expose_php = Off to symfony.prod.ini 

torbenbr:
https://www.php.net/manual/en/ini.core.php#ini.expose-php


maxhelias (approved):

maxhelias:
Good catch! Thanks @torbenbr

PR #118, opened 2021-03-23
fix: Caddy error message servers

gponty:
fix: error message : run: adapting config using caddyfile: /etc/caddy/Caddyfile:5: unrecognized global option: servers

https://caddyserver.com/docs/caddyfile/options


maxhelias (approved):

maxhelias (approved):

maxhelias:
Merci @gponty 😄

dunglas:
This should be reverted. The previous syntax was valid and the CI was green.

PR #119, opened 2021-03-23
feat: create database at startup

gponty:
feat: Try to create empty database at startup, in order to avoid the error message : "The database is not up or not reachable" if database doesnt exist.


maxhelias (approved):

maxhelias:
Thanks @gponty !

dunglas:
This should not be necessary. The DB is automatically created by the Postgres image. IMHO we should revert.

PR #128, opened 2021-04-08
feat: compat with MercureBundle 0.3

### LINKED ISSUE #127: Cannot Connect to mercure
Hello i follow over codes and documents and docker files but never find a solution to connect mercure in this structure

dunglas:
Fix #127.


docker-compose.yml
```diff
@@ -31,8 +31,9 @@ services:
       - php
     environment:
       SERVER_NAME: ${SERVER_NAME:-localhost, caddy:80}
-      MERCURE_PUBLISHER_JWT_KEY: ${MERCURE_PUBLISHER_JWT_KEY:-!ChangeMe!}
-      MERCURE_SUBSCRIBER_JWT_KEY: ${MERCURE_SUBSCRIBER_JWT_KEY:-!ChangeMe!}
+      MERCURE_URL: ${MERCURE_URL:-http://caddy/.well-known/mercure}
```
mtarld, line 34: I think that these three env var are related to the `php` service, not `caddy`

mtarld: Except for the `MERCURE_JWT_SECRET` which should stay in both services

mtarld, line 34: If going under the `php` service, we can't write `MERCURE_URL: ${MERCURE_URL:-http://caddy/.well-known/mercure}` because in that way, the value from the env var will be fetched from the `.env` file (which is not what we want because of the recipe)

mtarld (changes requested):

maxhelias:
@dunglas Do you have time to finish these?

dunglas:
@maxhelias it's on my todo list but it will not be before the 17th of May (I'm in vacation next week).

mtarld (approved):

PR #157, opened 2021-06-18
fix: download the correct skeleton in the Dockerfile

dunglas:


Dockerfile
```diff
@@ -94,7 +94,7 @@ ARG SYMFONY_VERSION=""
 ENV SYMFONY_VERSION ${SYMFONY_VERSION}
 
 # Download the Symfony skeleton and leverage Docker cache layers
-RUN composer create-project "symfony/skeleton ${SYMFONY_VERSION}" . --stability=$STABILITY --prefer-dist --no-dev --no-progress --no-interaction; \
+RUN composer create-project "${SKELETON} ${SYMFONY_VERSION}" . --stability=$STABILITY --prefer-dist --no-dev --no-progress --no-interaction; \
```
dunglas, line 97: Actually, I wonder why we need this. Couldn't this line be totally removed?

PR #180, opened 2021-10-03
Introduce a Makefile

COil:
Hi, in my projects I like to have a Makefile which provides some shortcuts. I thought It could be a good idea to ease the process even more for newcomers.

See you. COil


dunglas:
The problem is that Makefiles aren't cross-platform. Maybe could we add a doc entry explaining how to add a Makefile instead of shipping it directly in the repository?

maks-rafalko:
Just my 2 cents, to support @COil:

 `Makefile` is supported by MacOS and Linux. 

Getting rid of it just because of Windows seems strange, IMO.

If it's not supported on the system of the end user - it won't be used. For others - it will help and will be an expiration.

(or am I missing something?)

rowanparker:
IMHO these are a few reasons why you might not want to add this:

>  I thought It could be a good idea to ease the process even more for newcomers.

I think for ease of use for newcomers, the instructions shouldn't have two options (run this or that command). So if you did include the makefile, it's less confusing if you just present the docker commands only.

> Getting rid of it just because of Windows seems strange, IMO.

The point of using composer (both as developer and for deployment) is that it shouldn't matter what your host OS is. So everyone can just type `docker-compose up -d` and it "just works". A Windows user can't make use of these. You'd would inevitably have someone offer a Powershell cmdlet or a batch file. You'd then have additional files that are never used by some developers complicating the repository.

> If it's not supported on the system of the end user - it won't be used. For others - it will help and will be an expiration.

The fact that it is _not_ needed, only optional is a good reason to leave it out of a project that is intending to be used in production (one less vector for a bug).

> (or am I missing something?)

You're adding a new file that needs to be maintained, even though the subset of developers that would use is likely to be very small.

A docs page on how to implement a makefile is probably a better option.

COil:
I've made a new doc entry. I've put the template in a new `dist` folder, is it OK to put it there?

COil:
Done.

docs/makefile.md
```diff
@@ -0,0 +1,79 @@
...
+# Executables (local)
+DOCKER_COMP = docker-compose
+
+# Docker containers
+PHP_CONT = $(DOCKER_COMP) exec php
+
+# Executables
+PHP      = $(PHP_CONT) php
+COMPOSER = $(PHP_CONT) composer
+SYMFONY  = $(PHP_CONT) bin/console
```
bigindian66, line 35: Hello,
IMO it seems more logical to write :
```suggestion
SYMFONY  = $(PHP) bin/console
```
Moreover, in my case, I had the following error trying to use it as is :
`OCI runtime exec failed: exec failed: unable to start container process: exec: "bin/console": permission denied: unknown
make: *** [Makefile:50: sf] Error 126`
Thanks for reading this,
Bye.

bigindian66 (commented):
Thanks for your attention

docs/makefile.md
dunglas: LGTM. Would you mind opening a PR?

bigindian66: Sorry, I don't know GitHub very well, thought you could just accept the suggestion from here...
I've opened a PR as asked !

maxhelias: LGTM too, there is rather a problem elsewhere. The Symfony command must be executed inside the php container and not on the machine that does not have the same PHP context.

PR #181, opened 2021-10-13
fix: prevent installing another Mercure hub

dunglas:
Prevent the installation of https://github.com/symfony/recipes/pull/1007 because it's not needed in our case.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -54,7 +54,13 @@ services:
         published: 443
         protocol: udp
 
+# Mercure is installed as a Caddy module, prevent the Flex recipe from installing another service
+###> symfony/mercure-bundle ###
+###< symfony/mercure-bundle ###
+
 volumes:
   php_socket:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -54,7 +54,13 @@ services:
...
+# Mercure is installed as a Caddy module, prevent the Flex recipe from installing another service
+###> symfony/mercure-bundle ###
+###< symfony/mercure-bundle ###
+
 volumes:
   php_socket:
   caddy_data:
   caddy_config:
+###> symfony/mercure-bundle ###
+###< symfony/mercure-bundle ###
```
rogeriolino, line 66: hi @dunglas, is this duplicate placeholder expected?

maxhelias: @rogeriolino Yes, we need the marker to indicate to flex that the configuration is already present in services & volumes section.

PR #188, opened 2021-11-04
Fix: Do not run composer install in prod

mmenozzi:
IMHO always running composer install in PHP entrypoint is wrong.
In production this will:
* Install also dev dependencies which should not be installed
* Slow the starting phase of the container and this in turns increase the deployment downtime

So assume that in production the container runs from the image built with the Dockerfile, run composer install in entrypoint is wrong.

### DIFF
--- a/docker/php/docker-entrypoint.sh
+++ b/docker/php/docker-entrypoint.sh
@@ -26,11 +26,12 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		cd -
 
 		rm -Rf tmp/
-	elif [ "$APP_ENV" != 'prod' ]; then
-		rm -f .env.local.php
 	fi
 
-	composer install --prefer-dist --no-progress --no-interaction
+	if [ "$APP_ENV" != 'prod' ]; then
+		rm -f .env.local.php
+		composer install --prefer-dist --no-progress --no-interaction
+	fi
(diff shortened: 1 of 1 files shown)


dunglas (approved):
Maybe can this be merged with the previous elif (changing the `elif` by a `if`)?

mmenozzi:
> Maybe can this be merged with the previous elif (changing the `elif` by a `if`)?

You mean like the following?

```php
	# The first time volumes are mounted, the project needs to be recreated
	if [ ! -f composer.json ]; then
		CREATION=1
		composer create-project "$SKELETON $SYMFONY_VERSION" tmp --stability="$STABILITY" --prefer-dist --no-progress --no-interaction --no-install
		cd tmp
		composer config --json extra.symfony.docker 'true'
		cp -Rp . ..
		cd -
		rm -Rf tmp/
	fi

	if [ "$APP_ENV" != 'prod' ]; then
		rm -f .env.local.php
		composer install --prefer-dist --no-progress --no-interaction
	fi
```

If so, I agree. I don't get why there's an elif now. If it's not prod the `.env.local.php` should be removed regardless the existence of the `composer.json`.

Do you prefer this change in this PR or in a separate one?

dunglas:
Yes it's what I mean! You can do it in this PR.

mmenozzi:
Done @dunglas!

dunglas:
Thanks!

PR #190, opened 2021-11-12
Sync Docker and composer PHP versions in new projects

rowanparker:
At time of writing, if you clone and create a new project, the PHP container will use PHP 8.0, but the composer.json file will specify >=7.2.5, this makes them the same.


docker/php/docker-entrypoint.sh
```diff
@@ -21,6 +21,7 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		composer create-project "$SKELETON $SYMFONY_VERSION" tmp --stability="$STABILITY" --prefer-dist --no-progress --no-interaction --no-install
 
 		cd tmp
+		composer require "php:$PHP_VERSION"
```
maxhelias, line 24: With `php:>=$PHP_VERSION` would be better. Please, can you modify it?

rowanparker: Done

dunglas:
Thanks!

PR #199, opened 2021-12-15
Fix multistage alias from lower to uppercase

pietereggink:


docs/xdebug.md
```diff
@@ -32,7 +32,7 @@ As example, an override could look like this:
 version: "3.4"
 
 services:
-  php:
+  php-fpm:
```
maxhelias, line 35: The service name is not correct like this. See : https://github.com/dunglas/symfony-docker/blob/main/docker-compose.yml#L4

maxhelias (changes requested):

docs/xdebug.md
pietereggink: Sorry my bad, I'm using my own docker-compose.yml thats the reason. I will change it.

dunglas (approved):

maxhelias (approved):
Thanks!

PR #224, opened 2022-03-06
fix: check for migrations in subdirectories

qooplmao:
The entrypoint only check whether there are PHP files in the `./migrations` directory but they can be generated in subdirectories when using the `BY_YEAR` or `BY_YEAR_AND_MONTH` configurations.


maxhelias (approved):

maxhelias:
Thanks @qooplmao ! 😃

qooplmao:
Glad to be of use.



> I think you can make the same contribution to api-platform docker : https://github.com/api-platform/api-platform/blob/main/api/docker/php/docker-entrypoint.sh ;)

Done [#2153](https://github.com/api-platform/api-platform/pull/2153)

PR #227, opened 2022-03-12
Remove iconv workaround

o-alquimista:
This workaround is no longer needed because the problem [was fixed](https://github.com/docker-library/php/issues/240#issuecomment-1064627426) in the php image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,12 +17,7 @@ RUN apk add --no-cache \
 		file \
 		gettext \
 		git \
-		gnu-libiconv \
 	;
-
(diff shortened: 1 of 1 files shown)


maxhelias (approved):

maxhelias:
Thanks @o-alquimista, you can do the same change on the [api-platform docker](https://github.com/api-platform/api-platform/blob/main/api/Dockerfile) ?

o-alquimista:
> Thanks @o-alquimista, you can do the same change on the [api-platform docker](https://github.com/api-platform/api-platform/blob/main/api/Dockerfile) ?

Done: https://github.com/api-platform/api-platform/pull/2161

Issue #242, opened 2022-04-14
Slow on windows


Vandersteen:
A fresh install on windows with wsl2 gives me around 5-6 seconds for every call I make.

Do you have any suggestions on how to improve the performance on windows ?

I believe it to be IO issues but i'm not sure

stefanpoensgen:
I'm using wsl/wsl2 approx two years now and never experinced such an issue.
I would try to exclude wsl2 from windows defender realtime protection and if you use other protection software, go back to windows defender.

Ezar101:
@Vandersteen 

> To get the best out of the file system performance when bind-mounting files, we recommend storing source code and other data that is bind-mounted into Linux containers (i.e., with docker run -v <host-path>:<container-path>) in the Linux file system, rather than the Windows file system. You can also refer to the [recommendation](https://docs.microsoft.com/en-us/windows/wsl/compare-versions) from Microsoft.

[https://docs.docker.com/desktop/windows/wsl/#best-practices](https://docs.docker.com/desktop/windows/wsl/#best-practices)

aelfannir:
I was facing the same issue, enabled wsl2/disabled real-time protection didn't help

I've found that `vendor` and  `var` folders are affecting the performance

*docker-compose.yml*
```yml
...
services:
  php:
      ...
      volumes:
        - php_socket:/var/run/php
        - vendor:/srv/app/vendor
        - var:/srv/app/AE-API/var
        
volumes:
  ...
  vendor:
  var:
```

PR #270, opened 2022-07-31
Windows performance hint

robinbrisa:
Added commented lines in _docker-compose.override.yml_ about improving performance on Windows by using Docker volumes for the vendor and var directories.

Information found in issue #242.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -9,6 +9,11 @@ services:
       # If you develop on Mac you can remove the var/ directory from the bind-mount
       # for better performance by enabling the next line
       # - /srv/app/var
+      # If you develop on Windows you can greatly improve performance by using Docker
+      # volumes for the vendor and var directories by enabling the next 2 lines and 
+      # the "volumes" block at the bottom of this file
+      # - vendor:/srv/app/vendor
+      # - var:/srv/app/var
     environment:
       APP_ENV: dev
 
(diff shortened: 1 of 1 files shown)


docker-compose.override.yml
```diff
@@ -9,6 +9,11 @@ services:
       # If you develop on Mac you can remove the var/ directory from the bind-mount
       # for better performance by enabling the next line
       # - /srv/app/var
+      # If you develop on Windows you can greatly improve performance by using Docker
+      # volumes for the vendor and var directories by enabling the next 2 lines and 
+      # the "volumes" block at the bottom of this file
+      # - vendor:/srv/app/vendor
+      # - var:/srv/app/var
```
dunglas, line 16: Couldn't we use the same syntax as for Mac? This should have the same performance benefit.

docker-compose.override.yml
```diff
@@ -9,6 +9,11 @@ services:
       # If you develop on Mac you can remove the var/ directory from the bind-mount
       # for better performance by enabling the next line
       # - /srv/app/var
+      # If you develop on Windows you can greatly improve performance by using Docker
```
dunglas, line 12: This works for both Mac and Windows. I suggest merging both comments.

dunglas (changes requested):

robinbrisa:
Salut Kévin, you're right, using the same syntax works indeed. 
I just made some tests and the vendor dir needs to be added as it's the one which makes the most difference performance-wise. Not sure if it's the same on Mac though. 
I made a new version to reflect this, removing the volumes and adding the vendor directory to the bind-mount exclusion.

dunglas (approved):

dunglas:
Thanks! And yes excluding the vendors improves performance on Mac too, but the drawback is that it's harder to hack a dependency (when working with a private Packagist instance hosting company packages or when hacking an open source library).

PR #272, opened 2022-08-01
docs: trust certificate on Windows

robinbrisa:
Added information about how to add the caddy cert to the trusted certificates on Windows as this info was only provided for Linux & Mac.

_Note:_ I'm not using the same "docker cp $(docker compose ps -q caddy)" syntax as this does not work in the standard command prompt on Windows and as far as I know there is no equivalent.


dunglas (approved):

dunglas:
Thank you!

PR #283, opened 2022-09-07
fix: prevent skiping database intialisation

fLaVz:
Prevent skiping database initialisation when DATABASE_URL is not set in .env  but set in docker-compose environment variables

### DIFF
--- a/docker/php/docker-entrypoint.sh
+++ b/docker/php/docker-entrypoint.sh
@@ -26,7 +26,7 @@ if [ "$1" = 'php-fpm' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		composer install --prefer-dist --no-progress --no-interaction
 	fi
 
-	if grep -q ^DATABASE_URL= .env; then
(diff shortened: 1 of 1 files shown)


dunglas:
I'm not sure to understand the use case. If you don't use `DATABASE_URL`, then this env bar shouldn't be in `.env` either. Or am I missing something?

fLaVz:
That is very true. 
I must have misunderstood something because in my case, I had the .env var set with the wrong path but the environment var in docker-compose with the right path. The docker-compose environment var did not override the .env file var. Once I figured this out, I tried to only utilize the var set in docker-compose which can't work due to the condition in the entrypoint.sh. But I assume, according to your answer, that this case should never happen and so this pull request.
Am I right ?

dunglas:
> The docker-compose environment var did not override the .env file var.

That's weird. The "real" env var should always take precedence on the one define in `.env`

fLaVz:
Yes, that is what I read, that's why I tried both without setting the var and with setting the right path in the .env. Both worked without the ".env" condition in the entrypoint.sh. I assume that I should have misconfigured my project somewhere then. Anyway, I feel like this pull request should not be.

PR #285, opened 2022-09-14
correct wrongly typed docker-compose

fullbl:
There is an error in documentation, docker-compose is a unique command!

### DIFF
--- a/docs/existing-project.md
+++ b/docs/existing-project.md
@@ -24,10 +24,10 @@ Double-check the changes, revert the changes that you don't want to keep:
 
 Build the Docker images:
 
-    docker compose build --no-cache --pull
+    docker-compose build --no-cache --pull
 
 Start the project!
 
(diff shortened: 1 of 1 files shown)


jdreesen:
Nope, that's the legacy docker compose V1. Docker compose V2 is `docker compose` instead of `docker-compose`, so the docs are correct.

senzowayne:
J'ai fais l'installation sur un ubuntu dernière version et c'est bien `docker compose` et plus `docker-compose`

fullbl:
My bad, I thought it was a typo. Thank you for the information, didn't know about it (https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command)!

maxhelias:
This is the new command with the space, see : https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command

PR #291, opened 2022-09-21
fix: stop use obsolete experimental_http3 option

bkosun:
> Caddy now enables [RFC 9114](https://datatracker.ietf.org/doc/rfc9114/)-compliant HTTP/3 by default. The experimental_http3 option has graduated and been removed.

https://github.com/caddyserver/caddy/releases/tag/v2.6.0

### DIFF
--- a/docker/caddy/Caddyfile
+++ b/docker/caddy/Caddyfile
@@ -1,12 +1,6 @@
 {
     # Debug
     {$DEBUG}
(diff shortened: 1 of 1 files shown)


maxhelias (approved):

maxhelias:
Thanks @bkosun ! 😃

Issue #276, opened 2022-08-05
Simplify php extensions installation process


vonalbert:
Since php extensions installation process is not easy and adding more of them can clutter your dockerfile making it unreadable, usually I use this script to do this task

https://github.com/mlocati/docker-php-extension-installer

This script installs and builds the extensions you need and their dependencies in a single apt/apk command and as a bonus it is both alpine and debian compatible. All you need to do is add the script to the dockerfile and execute it with the list of the required extensions.

Would you be interested in a PR?

dunglas:
I'm hesitating. This script is nice and could simplify the Dockerfile, but on the other hand we try to use as few external dependencies as possible and to be as close as possible from base images.

WDYT @maxhelias?

maxhelias:
Making the Dockerfile simpler/flexible is worth taking especially for installing php extensions as long as they are low impact.
This script would allow us to add a link to the list of easy-to-install extensions to make the project more accessible to people who may not be familiar with that aspect (php extension / docker) and add some value to the project without too much impact (I hope).
But before, I would like to see a try of this

dunglas:
Let's do this then. This script is reference in the official readme by the way.

vonalbert:
It was not referenced in the php image readme back then. Good thing now it is. Anyway I could take care of it if you want.

bogdaniel:
Late to the party but this addition to the Dockerfile seems super good even if it's adding an extra external dependency. Extensions are now easier to install :)

PR #298, opened 2022-10-04
Use php extensions installer

b1rdex:
https://github.com/mlocati/docker-php-extension-installer

Fixes https://github.com/dunglas/symfony-docker/issues/276

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,10 @@ ENV APP_ENV=prod
 
 WORKDIR /srv/app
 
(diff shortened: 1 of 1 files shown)


dunglas (approved):
I like it!

I wonder if we couldn't provide a better integration with Flex thanks to this script. For instance we could just list the extensions to install in the flex recipe and pass the list to this script.

maxhelias (approved):
I like it too!

maxhelias:
Thanks @b1rdex ! :)

PR #331, opened 2022-11-24
chore: remove build args

dunglas:
Remove build args (as official images) and use the new installation method for `mlocati/php-extension-installer`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ ENV APP_ENV=prod
 WORKDIR /srv/app
 
 # php extensions installer: https://github.com/mlocati/docker-php-extension-installer
-COPY --from=mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/local/bin/
+COPY --from=mlocati/php-extension-installer --link /usr/bin/install-php-extensions /usr/local/bin/
 
 # persistent / runtime deps
 RUN apk add --no-cache \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -24,8 +20,7 @@ ENV APP_ENV=prod
 WORKDIR /srv/app
 
 # php extensions installer: https://github.com/mlocati/docker-php-extension-installer
-ADD https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions /usr/local/bin/
-RUN chmod +x /usr/local/bin/install-php-extensions
+COPY --from=mlocati/php-extension-installer /usr/bin/install-php-extensions /usr/local/bin/
```
maxhelias, line 23:
```suggestion
COPY --from=mlocati/php-extension-installer --link /usr/bin/install-php-extensions /usr/local/bin/
```

maxhelias (approved):
Thanks @dunglas

thePanz:
So, supporting an "older" version of PHP would need more work?

Would you provide an example how to achieve it?

maxhelias:
What do you mean ? You can always change the PHP version as before. Currently, how do you override the argument ?

thePanz:
I use a Makefile with the following:

```
SERVER_NAME="blabla-bla.localhost"
PHP_VERSION="7.4"
PROJECT_NAME="blabla.bla"
CLI_PARAMS="--project-name=$(PROJECT_NAME)"

.PHONY: build
build:
    SERVER_NAME="${SERVER_NAME}" docker-compose $(CLI_PARAMS) build --build-arg PHP_VERSION=${PHP_VERSION}

maxhelias:
Indeed, your use case is a good one for rollback for me. What do you think @dunglas ?

PR #339, opened 2022-12-01
docs: use correct .PHONY syntax with colon (:)

wira:
Current makefile template in docs/makefile.md is using equal sign when declaring .PHONY, this will be ignored because supposedly the correct syntax is using colon (:).

makefile-phony


maxhelias (approved):

maxhelias:
Thanks for the contribution!

PR #352, opened 2022-12-21
Allow override PHP version with build arg

maxhelias:
See @thePanz comment : https://github.com/dunglas/symfony-docker/pull/331#issuecomment-1327599136


dunglas:
I'm not sure, this kind of trick cause more harm than good IMHO. It's still possible to just edit the local copy of Dockerfile.

PR #353, opened 2022-12-21
Use hadolint to check dockerfile

maxhelias:
It's a try. We need https://github.com/hadolint/hadolint/issues/826 to get a green CI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,20 @@
 # https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage
 # https://docs.docker.com/compose/compose-file/#target
 
+# Builder images
+FROM composer/composer:2-bin AS composer
+
+FROM mlocati/php-extension-installer:latest AS php_extension_installer
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -64,18 +68,18 @@ CMD ["php-fpm"]
 ENV COMPOSER_ALLOW_SUPERUSER=1
 ENV PATH="${PATH}:/root/.composer/vendor/bin"
 
-COPY --from=composer/composer:2-bin --link /composer /usr/bin/composer
+COPY --from=composer --link /composer /usr/bin/composer
 
 # prevent the reinstallation of vendors at every changes in the source code
-COPY composer.* symfony.* ./
+COPY --link composer.* symfony.* ./
```
dunglas, line 74: Is this safe? Will it not prevent creating a specific cache layer just with the vendors?

maxhelias: I think, according to what is indicated in the docker documentation, it seems to be : https://docs.docker.com/engine/reference/builder/#benefits-of-using---link

dunglas (approved):
LGTM!

PR #355, opened 2022-12-21
Match / and . in Makefile help regex

KajdeMunter:
It is common practice to group Makefile targets using `.` and `/`. 

For example:
`make test.unit` or `make test/unit`

This change makes it so that those commands will be included in the help target.


maxhelias (approved):

maxhelias:
Thanks @KajdeMunter !

PR #368, opened 2023-01-19
docs: add server name tips and custom https port note

### LINKED ISSUE #357: Documentation about digital certificates: mention Docker Desktop to inspect ENV variables
I'm setting up a new project using this configuration as skeleton.

Once built the containers and copied the cert from the Caddy's container to my local machine, connecting to myapp.localhost gave an error "ERR_SSL_PROTOCOL_ERROR".

Going around to check the issue, I, in the end, discovered I wrongly set the `SERVER_NAME` env variable: in fact, I didn't added it to the `.env` file and so the `docker compose up -d` command didn't set it.

To discover this issue I checked the configuration of the container through Docker Desktop (click on the container name, then click on "Inspect"): there I discovered the `SERVER_NAME` env variable was wrongly set.

This information may be very useful to the ones who are trying to debug issues with TLS certificates and I think it could be added to the documentation.

Checking the [Troubleshooting](https://github.com/dunglas/symfony-docker/blob/main/docs/troubleshooting.md) page, however, I saw it reports back to [TLS page](https://github.com/dunglas/symfony-docker/blob/main/docs/tls.md): in this page, however, there is no help to troubleshoot issues that may happen.

So, the question is: where to add this bit of information about the use of Docker Desktop (or, at least, to double-check the `SERVER_NAME` env var is correctly set)?

maxhelias:
Fix #357

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -36,7 +36,7 @@ Use the `SERVER_NAME` environment variable to define your custom server name(s).
 
 If you use Mercure, keep `caddy:80` in the list to allow the PHP container to request the caddy service.
 
-*Tips : You can define your server name variable in your `.env` file to keep it at each up*
+*Tips: You can define your server name variable in your `.env` file to keep it at each up*
 
 ## Using custom HTTP ports
 
--- a/docs/build.md
+++ b/docs/build.md
@@ -46,4 +46,4 @@ Use the environment variables `HTTP_PORT`, `HTTPS_PORT` and/or `HTTP3_PORT` to a
 
 to access your appplication on [https://localhost:4443](https://localhost:4443).
 
-*Note : Let's Encrypt ony supports the standard HTTP and HTTPS ports. Creating a LE certificate for another port will not work, you have to use the standard ports or to configure Caddy to use another provider.*
+*Note: Let's Encrypt only supports the standard HTTP and HTTPS ports. Creating a Let's Encrypt certificate for another port will not work, you have to use the standard ports or to configure Caddy to use another provider.*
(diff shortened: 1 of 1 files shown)


docs/build.md
```diff
@@ -36,10 +36,14 @@ Use the `SERVER_NAME` environment variable to define your custom server name(s).
 
 If you use Mercure, keep `caddy:80` in the list to allow the PHP container to request the caddy service.
 
+*Tips : You can define your server name variable in your `.env` file to keep it at each up*
```
dunglas, line 39:
```suggestion
*Tips: You can define your server name variable in your `.env` file to keep it at each up*
```

docs/build.md
```diff
@@ -36,10 +36,14 @@ Use the `SERVER_NAME` environment variable to define your custom server name(s).
...
+
 ## Using custom HTTP ports
 
 Use the environment variables `HTTP_PORT`, `HTTPS_PORT` and/or `HTTP3_PORT` to adjust the ports to your needs, e.g.
 
     HTTP_PORT=8000 HTTPS_PORT=4443 HTTP3_PORT=4443 docker compose up --build
 
 to access your appplication on [https://localhost:4443](https://localhost:4443).
+
+*Note : Let's Encrypt ony supports the standard HTTP and HTTPS ports. Creating a LE certificate for another port will not work, you have to use the standard ports or to configure Caddy to use another provider.*
```
dunglas, line 49:
```suggestion
*Note: Let's Encrypt only supports the standard HTTP and HTTPS ports. Creating a Let's Encrypt certificate for another port will not work, you have to use the standard ports or to configure Caddy to use another provider.*
```

dunglas (changes requested):

PR #371, opened 2023-01-26
Update docker-entrypoint.sh

rrotteveel:
On my machine the php instance was not coming up[ because it was searching for the database. Finally I logged in to the terminal of the image and there I was able to execute the command 'php bin/console dbal:run-sql "SELECT 1"'. Hence, I believe php is missing from the test in the 'until' on line 40.


7-zete-7:
Will be sufficient to make `bin/console` file executable with `chmod +x bin/console` on your machine, and the accompanying change will not be necessary.

Issue #377, opened 2023-02-07
Use heath check feature from Caddy


maidmaid:
Caddy has its own [health check feature](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#health_uri). In the next Caddy [2.6.3 version](https://github.com/caddyserver/caddy/issues/5281), the heath check will also work with `php-fpm.sock` and it will be possible to write something like :

```Caddyfile
localhost

route {
    root * /srv/app/public
    php_fastcgi unix//var/run/php/php-fpm.sock {
        health_uri /ping
    }
}
```

The advantage compared to the current [docker-healthcheck.sh](https://github.com/dunglas/symfony-docker/blob/main/docker/php/docker-healthcheck.sh) is that Caddy is really waiting php-fpm before serving traffic to it. It's very important when we build only one container with php and caddy together as in this described [here](https://github.com/dunglas/symfony-docker/issues/375).

maidmaid:
[2.6.3 released](https://github.com/caddyserver/caddy/releases/tag/v2.6.3) :tada:

maxhelias:
Close has explain here : https://github.com/dunglas/symfony-docker/pull/384#issuecomment-1658311145

PR #384, opened 2023-02-27
Use native caddy healthcheck

maidmaid:
Resolves https://github.com/dunglas/symfony-docker/issues/377

```
caddy_1  | {"level":"info","ts":1677492269.4432561,"logger":"http.handlers.reverse_proxy.health_checker.active","msg":"HTTP request failed","host":"localhost","error":"Get \"http://localhost/ping\": dialing backend: dial unix /var/run/php/php-fpm.sock: connect: no such file or directory"}
php_1    | Executing script assets:install public [OK]
php_1    | [27-Feb-2023 10:04:29] NOTICE: fpm is running, pid 1
php_1    | [27-Feb-2023 10:04:29] NOTICE: ready to handle connections
caddy_1  | {"level":"info","ts":1677492270.444267,"logger":"http.handlers.reverse_proxy.health_checker.active","msg":"host is up","host":"localhost"}
```

### DIFF
--- a/docker/php/php-fpm.d/zz-docker.conf
+++ b/docker/php/php-fpm.d/zz-docker.conf
@@ -6,3 +6,4 @@ process_control_timeout = 20
 listen = /var/run/php/php-fpm.sock
 listen.mode = 0666
 ping.path = /ping
+access.suppress_path[] = /ping
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,3 +40,6 @@ jobs:
               run: curl http://localhost
             - name: Check HTTPS reachability
               run: curl -k https://localhost
+            - name: Log
+              if: always()
+              run: docker-compose logs
--- a/docker/caddy/Caddyfile
+++ b/docker/caddy/Caddyfile
@@ -26,7 +26,11 @@ route {
         {$MERCURE_EXTRA_DIRECTIVES}
     }
     vulcain
-    php_fastcgi unix//var/run/php/php-fpm.sock
+    php_fastcgi unix//var/run/php/php-fpm.sock {
+        health_uri /ping
+        health_interval 1s
+        lb_try_duration 5s
+    }
     encode zstd gzip
     file_server
 }


docker/php/php-fpm.d/zz-docker.conf
```diff
@@ -6,3 +6,4 @@ process_control_timeout = 20
 listen = /var/run/php/php-fpm.sock
 listen.mode = 0666
 ping.path = /ping
+access.suppress_path[] = /ping
```
maidmaid, line 9: See https://github.com/php/php-src/commit/327bb219867b16e1161da632fba170f46692484b

bzoks:
But this is then not health-check run or understood by docker, for it to restart php service when healthcheck fails... this would just "mark" php-fpm as being down for caddy and nothing else... no auto-heal with restart... 
Second part (hiding from logs) is great :)

maidmaid:
Thanks for the feedback @bzoks. I simplified the PR by keeping the docker healthcheck as-it.

dunglas:
What's the benefit of using Caddy's health checks in addition to Docker / K8S health checks? This will multiply the number of check requests for nothing as load balancing is already handled by Docker (or K8S).

+1 for hiding the ping requests for the logs.

maidmaid:
I had to deploy my container in the serverless Google Cloud Run solution and the health checks defined in the Dockerfile are not used. So, the traffic was served whereas php fpm wasn't ready. Using the Caddy health check could be a good last safety net imo.

dunglas:
This looks like a bug or limitation in Google Cloud Run. IMHO it would be better to add docs explaining how to configure this project to be compatible with Google Cloud Run than increasing the load on every other standard deployments (and making the project more complex). WDYT?

maidmaid:
I personally like to have health checks to container level and also to application level for a better faults tolerance in production. They are complementary imo. However, I totally understand you want to keep project source simple. I'll update the PR soon.

dunglas:
Thanks.

Having both will cause issues. For instance, if you use a Kubernetes service but have a Caddy health check, the domain associated with the service (which is a load balancer) could be marked by Caddy as unhealthy why it shouldn't (because K8S will manage that).

ihmels:
> I had to deploy my container in the serverless Google Cloud Run solution and the health checks defined in the Dockerfile are not used. So, the traffic was served whereas php fpm wasn't ready. Using the Caddy health check could be a good last safety net imo.

Google Cloud Run uses Knative which in turn uses Kubernetes and Kubernetes doesn't use Docker healthchecks but has its [own mechansims](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/).

dunglas:
Closing as won't fix. However, don't hesitate to open a docs PR explaining how to use native Caddy health checks for GCR, Knative etc.

PR #391, opened 2023-03-27
docs: add tips to pass local environment variables to containers

ker0x:


docs/production.md
```diff
@@ -102,3 +102,18 @@ As a shortcut, `private_ranges` may be configured to trust all private IP ranges
...
+By default, `.env.local` and `.env.*.local` files are excluded from production images.
+If you want to pass them to your containers, you can use the [`env_file` attribute](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute):
+
+```yaml
+# docker-compose.prod.yml
+
+services:
+    php:
+        env_file:
+            - .env.local
```
maxhelias, line 117: I am not convinced that this is a good practice for production.

In the case of Symfony application, there is the dump-env: https://github.com/dunglas/symfony-docker/blob/cb0f62eb31dd35c856af58928f68c93386bc674f/Dockerfile#L99

And in the other cases, it would be better to pass the desired variables via the `environment` section

ker0x: `env_file` as the same behavior as the `environment` attribute, it's juste a shortcut to pass a file instead of listing all environment variables in `docker-compose.yml`. It will **not** copy the `.env.local` file inside the container.

`composer dump-env prod` will only dump environment variables from `.env` file since `.env.local` is exclude from the image through `.dockerignore`

jameskitt616:
Is there any progress on this? It would be great to have `.env.local` working.

ker0x:
@maxhelias Done 😉 !

maxhelias (approved):

maxhelias:
Thanks !

PR #417, opened 2023-06-13
Upgrade Postgres default version to 15

WebMamba:
In the 2.8 doctrine bundle recipes the default Postgres version is 15 so let's keep in sync with it.

See: https://github.com/symfony/recipes/blob/356ad033b1740ad32c780d050c2bf0ee128a3ca4/doctrine/doctrine-bundle/2.8/manifest.json#LL16C87-L16C89


dunglas:
Thanks!

PR #186, opened 2021-10-31
Document how to add npm (or yarn)

rowanparker:
As per https://github.com/dunglas/symfony-docker/issues/129

PR #425, opened 2023-07-02
Document how to add Webpack Encore

packagespace:
Cont. #186 

As explained in the last thread, this config works in a local env with HMR and in prod. No problem with Symfony UX packages either. Added a few improvements to avoid reinstalling node modules at every change in the source code.

I was wondering if there is a way this could be handled by a flex recipe, but if the only additions in the `Dockerfile` have to be made in the `app_php` image I'm not sure how it could be done.

Some extra notes in case it could help anyone tackle this / improve my approach: 
 - Build output must be copied to both php image for `entrypoints.json` and caddy image for assets serving
 - Node container for dev environment does not need to be built from symfony_node image except for purposes of keeping the same node version
 - Some of the encore options might be deprecated (see https://github.com/symfony/webpack-encore/pull/1095)


jmarcone:
Hi , I tried to use your solution , but I can't build the docker image:

image

Do you have an idea of what may be happening?

Thanks!

ziat1988:
Hello, 
Any update from this PR? I'm trying integrate Symfony UX and don't know how to do it correctly . The repo don't include nodejs container or any document related

clemcoder:
I'm in the same situation. Do you have any idea when the addition of the asset build will be integrated into the main branch? Thanks.

cptwtf:
Also failing to get this to work with the current state of the repo.
Anyone who can provide some help?

astronati:
Any news about this enhancement?

zpi12lmm:
If you are looking for a simple and quick solution, you can do it as follows: https://github.com/dunglas/symfony-docker/issues/497

devnix:
I'm finding this closed pull request very useful indeed.

I like the idea of AssetMapper, but 

1. it's not a viable alternative for a lot of projects
2. is a valid option as good as Encore
3. I wish this repository would not be used as an indoctrination tool

PR #435, opened 2023-08-03
docs: document php-fpm path exclusion from access logs

thePanz:


dunglas:
Thanks for the PR, but we try to keep these files as small as possible. Maybe just a comment like "expose a URL for health checks"?

Users should refer to FPM docs for more details.

thePanz:
@dunglas changed to `Exclude the /ping path from access logs (directive available for PHPv8.2)`

WDYT?

ps: please squash the commits from this PR :)

docker/php/php-fpm.d/zz-docker.conf
```diff
@@ -6,4 +6,6 @@ process_control_timeout = 20
 listen = /var/run/php/php-fpm.sock
 listen.mode = 0666
 ping.path = /ping
+
+; Exclude the /ping path from access logs (directive available for PHPv8.2)
```
maxhelias, line 10:
```suggestion
; Exclude the /ping path from access logs (directive available since PHP 8.2)
```

dunglas:
I stil doubt this comment is worth it.

thePanz:
It helps if a user is personalizing the docker image (example: using PHP v8.1) and would help to quickly identify the failure of starting PHP-FPM.

Why a user would need to personalize the docker image with an older version of PHP?
- users might use the docker image for local development, while keeping the specs similar to the staging/production deployments which do not use docker

dunglas:
PHP 8.1 will be EOLed on 25 Nov 2023, this is a very temporary patch :)

thePanz:
True, but do not forget about all of those *legacy* projects that can not upgrade ;-)

FYI: https://packagist.org/packages/monolog/monolog/php-stats?ref=blog.packagist.com php 7.x is still out there, running on 35% of the monolog installs (monolog is just an example) :)

maxhelias:
With the rework of the template for FrankenPHP, this file no longer exists.

PR #441, opened 2023-08-08
Add test command to healthcheck in docker-compose.yml

gabrielpenide:
According to Compose docs, the `healthcheck` instruction should specify a `test` command in `docker-compose.yml`.

See: https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck

PS: Also formatting some spaces to match the rest of the `Dockerfile`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,8 +73,8 @@ VOLUME /srv/app/var/
 RUN mv "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"
 
 RUN set -eux; \
-	install-php-extensions \
-    	xdebug \
+    install-php-extensions \
+		xdebug \
     ;
 
 COPY --link docker/php/conf.d/app.dev.ini $PHP_INI_DIR/conf.d/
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -73,8 +73,8 @@ VOLUME /srv/app/var/
 RUN mv "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"
 
 RUN set -eux; \
-	install-php-extensions \
-    	xdebug \
+    install-php-extensions \
+		xdebug \
```
jurchiks, line 77: well this is obviously an unnecessary change

dunglas:
This isn't necessary because we have a default health check: https://github.com/dunglas/symfony-docker/blob/main/Dockerfile#L52

Actually, we could even delete the whole `healthcheck` session of `docker-compose.yml`.

gabrielpenide:
Fair enough. I was in fact wondering if that lines in both files would be redundant.

Thanks.

PR #449, opened 2023-08-17
Small update for production deployment documentation

latysh:
When building for production we need to specify the prod compose file specifically otherwise there are issues. I hope it helps other people struggling with the same issue.


dunglas:
This shouldn't be necessary unless local images already exist. In this case, passing `--build` to `docker compose up` should be enough.

PR #450, opened 2023-08-18
Set chmod in COPY instructions for docker scripts

matiboux:
Use `COPY --chmod` instead of `COPY` + `RUN chmod`.
This removes two layers and reduce overall image size.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,13 +46,11 @@ COPY --link docker/php/conf.d/app.ini $PHP_INI_DIR/conf.d/
 COPY --link docker/php/php-fpm.d/zz-docker.conf /usr/local/etc/php-fpm.d/zz-docker.conf
 RUN mkdir -p /var/run/php
 
-COPY --link docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
-RUN chmod +x /usr/local/bin/docker-healthcheck
+COPY --link --chmod=+x docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
 
 HEALTHCHECK --interval=10s --timeout=3s --retries=3 CMD ["docker-healthcheck"]
 
-COPY --link docker/php/docker-entrypoint.sh /usr/local/bin/docker-entrypoint
-RUN chmod +x /usr/local/bin/docker-entrypoint
+COPY --link --chmod=+x docker/php/docker-entrypoint.sh /usr/local/bin/docker-entrypoint
 
 ENTRYPOINT ["docker-entrypoint"]
 CMD ["php-fpm"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -46,13 +46,11 @@ COPY --link docker/php/conf.d/app.ini $PHP_INI_DIR/conf.d/
 COPY --link docker/php/php-fpm.d/zz-docker.conf /usr/local/etc/php-fpm.d/zz-docker.conf
 RUN mkdir -p /var/run/php
 
-COPY --link docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
-RUN chmod +x /usr/local/bin/docker-healthcheck
+COPY --link --chmod=+x docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
```
dunglas, line 49: Good idea! However only octal notation is currently supported, `+x` will not work, this should be `500`.

dunglas (changes requested):

Dockerfile
matiboux: Oh that's disappointing. `--chmod=+x` fails silently, which led me to believe it worked.

I'm concerned about `--chmod=500`, is it possible it would cause ownership issues?
Would one of these be preferable?
- `--chmod=555`
- `--chmod=755` (git perms for an executable file)

dunglas: What is the current permission? Let's use the same so we're sure that we break nothing!

matiboux: Git only tracks if a file is executable or not, which results in files being either `644` or `755`.
Currently, `docker-healthcheck.sh` is indexed as an executable file, so it should checkout with `755`.
Let's use this then!

matiboux:
```suggestion
COPY --link --chmod=755 docker/php/docker-healthcheck.sh /usr/local/bin/docker-healthcheck
```

Dockerfile
```diff
@@ -46,13 +46,11 @@ COPY --link docker/php/conf.d/app.ini $PHP_INI_DIR/conf.d/
...
 
 HEALTHCHECK --interval=10s --timeout=3s --retries=3 CMD ["docker-healthcheck"]
 
-COPY --link docker/php/docker-entrypoint.sh /usr/local/bin/docker-entrypoint
-RUN chmod +x /usr/local/bin/docker-entrypoint
+COPY --link --chmod=+x docker/php/docker-entrypoint.sh /usr/local/bin/docker-entrypoint
```
matiboux, line 53:
```suggestion
COPY --link --chmod=755 docker/php/docker-entrypoint.sh /usr/local/bin/docker-entrypoint
```

dunglas (approved):

dunglas:
Thank you @matiboux

PR #458, opened 2023-08-28
update dependency 

fkhair:
Updated current dependency and remove glitch


maxhelias:
We don't need to publish images in a registry, we simply offer a template. Everyone is responsible for containerizing their own project.

PR #460, opened 2023-09-13
feat: switch to FrankenPHP

dunglas:
Use [FrankenPHP](https://frankenphp.dev) 🎉


bzoks:
Wouldn't it be "better" to add this as an alternative/parallel option (for example in a separate Dockerfile or at least a separate stage), besides php-fpm, especially because of this part in the readme: 

> FrankenPHP is very experimental, don't use it in production yet

Starting to use dockerized PHP can be a large step for newcomers, but in parallel going away from well-known php-fpm... could be a showstopper for some (too adventurous)

docker/frankenphp/Caddyfile
jdreesen, line 1: This file uses now a mix of spaces and tabs for indentation.

dunglas:
@bzoks I plan to create an `fpm` branch to still support FPM for a while. However, in the long term, the goal is to switch to FrankenPHP because it has been designed especially for this kind of use case. It is slimmer, easier to use, and faster than FPM.

Dockerfile
```diff
@@ -1,21 +1,19 @@
...
 
 
 # The different stages of this Dockerfile are meant to be built into separate images
 # https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage
 # https://docs.docker.com/compose/compose-file/#target
 
 
-# Base PHP image
-FROM php_upstream AS php_base
+# Base FrnakenPHP image
```
senzowayne, line 13:
```suggestion
# Base FrankenPHP image
```

PR #471, opened 2023-09-26
fix: session support

### LINKED ISSUE #470: Session cannot be set since upgrade to FrankenPHP
I tried to implement a User Login form from scratch following symfony documentation but I never got logged in properly. Debugging lead to the following results:
* I don't have a symfony session id ($requestStack->getSession()->getId() is always null
* I can see a session id in my browsers dev toolbar (Application Tab)
* The User provider properly loads the user but a call to $this->getUser() in any controller always gives null

What i can tell besides:
* The /var folder is not syncronized to local
* If I switch to the session native file session handler, the directory /var/sessions/dev is created but no file is stored
* I have an insecure connection but setting cookie secure to false does not help anyways

I would appreciate any idea what could cause this behaviour. Unfortunately I suspect something with the newly introduced FrankenPHP setup. The exact same approach always worked nicely before.

dunglas:
Closes #470

### DIFF
--- a/frankenphp/conf.d/app.ini
+++ b/frankenphp/conf.d/app.ini
@@ -1,4 +1,4 @@
-variables_order = EGPS
+variables_order = EGPCS
 expose_php = 0
 date.timezone = UTC
 apc.enable_cli = 1


maxhelias:
Thanks @dunglas for all the work on FrankenPHP, I'll be looking it in detail soon 😃

PR #478, opened 2023-10-02
Adding --detach option to the command

### LINKED ISSUE #436: `docker compose up --wait` does not detach, `-d` required
The latest commit replaced `-d`/`--detach` with `--wait`, however it does not seem to work the way the documentation says it should, at least not on Windows.

According to Docker's own documentation (https://docs.docker.com/engine/reference/commandline/compose_up/):

>`--wait`		Wait for services to be running|healthy. ***Implies detached mode.***

One might think this means that it runs the services, waits for them to be fully up, and THEN detaches _implicitly_, without requiring an explicit `-d` flag. However, if I run it with just  `--wait`, it does not detach. One might say the documentation is wrong or incorrectly worded - wouldn't be the first time -, or maybe it's broken on Windows (I had Docker Desktop 4.21.1, updated to 4.22.0, still the same).

But then I found this article: https://maciejwalkowiak.com/blog/docker-compose-waiting-containers-ready/

And if I use `--detach --wait`, then it works as it should, so the code and documentation changes from the latest commit should be fixed.

Annoyingly, I tried clicking on "Edit this page" in the Docker documentation, but it lead me nowhere, and searching the repo it opened for `Implies detached mode` gave me no results, so I can't fix the wording myself.

maxhelias:
Fix #436

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ with [FrankenPHP](https://frankenphp.dev) and [Caddy](https://caddyserver.com/) 
 
 1. If not already done, [install Docker Compose](https://docs.docker.com/compose/install/) (v2.10+)
 2. Run `docker compose build --no-cache` to build fresh images
-3. Run `docker compose up --pull --wait` to start the project
+3. Run `docker compose up --pull -d --wait` to start the project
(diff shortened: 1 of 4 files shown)


dunglas (approved):
I would use the long form for consistency, but LGTM!

PR #481, opened 2023-10-04
Replace spaces indentations by tabs indentations in dockerfile

BigDaddySigma:


dunglas (approved):
Good catch. It would be nice to find a linter for that.

stefantalen:
> Good catch. It would be nice to find a linter for that.

Combining [hadolint](https://github.com/hadolint/hadolint) with [hindent](https://github.com/mihaimaruseac/hindent)?

dunglas:
Isn't hindent a tool to indent Haskell files?

PR #487, opened 2023-10-16
Update makefile template

emmanuel-tilleuls:
I've added the `test` target in the makefile template in order to remember to add the `APP_ENV=test` before the `phpunit` command.
I'm not sure to use the correct letter : `c` for adding options to `phpunit` : 
- pros : the same letter is used for `composer` and `sf` targets
- cons : `o` would be more appropriate since it passes options to `phpunit` and not command

### DIFF
--- a/docs/makefile.md
+++ b/docs/makefile.md
@@ -47,7 +47,7 @@ SYMFONY  = $(PHP) bin/console
 
 # Misc
 .DEFAULT_GOAL = help
-.PHONY        : help build up start down logs sh composer vendor sf cc
+.PHONY        : help build up start down logs sh composer vendor sf cc test
 
 ## —— 🎵 🐳 The Symfony Docker Makefile 🐳 🎵 ——————————————————————————————————
 help: ## Outputs this help screen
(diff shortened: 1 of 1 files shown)


dunglas:
Isn't the env var set automatically by PHPUnit itself (in `phpunit.xml.dist`)?

emmanuel-tilleuls:
> Isn't the env var set automatically by PHPUnit itself (in `phpunit.xml.dist`)?

in `phpunit.xml.dist` APP_ENV is set for the `$_SERVER` superglobal but not for `$_ENV`

I wasn't able to create the «test» container inside a test with `KernelTestCase::getContainer` if I omit to set the APP_ENV correctly.
Another way to make this work is to add `<env name="APP_ENV" value="test" force="true" />` in `phpunit.xml.dist`

bzoks:
I recently noticed a change to `variables_order` (in scope of switching to FrankenPhp), which is now different than default/official PHP docker image, which has this:
`
variables_order = "GPCS"
`
Adding E to this list might change the behaviour of many libraries... perhaps the default behaviour of PHPUnit is also affected by this.

maxhelias (approved):

maxhelias:
@dunglas I like the idea of having a test command like these for the makefile. The use of the APP_ENV variable remains to be discussed, I imagine ?

dunglas (approved):

maxhelias:
Thanks @emmanuel-tilleuls

PR #491, opened 2023-10-23
fix: install Composer dependencies if vendor directory is empty

### LINKED ISSUE #490: Composer dependency installation condition fails
After merging PR #466, the way we check whether Composer dependencies need to be installed has changed:

```diff
-if [ "$APP_ENV" != 'prod' ]; then
+if [ ! -d 'vendor/' ]; then
```

https://github.com/dunglas/symfony-docker/pull/466/files#diff-14153d14d3f68eac82560b3b981b743970163e3732f87fd8d8b8b9a677429982L30

Please note that in case of using some recommendations to improve productivity:

https://github.com/dunglas/symfony-docker/blob/97c48c196b31efcb4a45106217fe3312ad9a9d80/docker-compose.override.yml#L13-L15

Condition `if [ ! -d 'vendor/' ]; then` will never be executed because the `vendor` directory will be created automatically before the `docker-entrypoint.sh` script is executed

Installing dependencies must be done if the `vendor` directory does not exist or is empty. In this case, you need to use a different condition. For example, you can use the following method:
```
	if [ -z "$(ls -A 'vendor/')" ]; then
		composer install --prefer-dist --no-progress --no-interaction
	fi
```

This condition will be true if the `vendor` directory does not exist (in which case the `ls` command will return an error and the command output will be empty) or if there are no files or subdirectories in the directory (in which case `ls -A` will output empty lines, and the output of the command will also be empty)

bkosun:
Fixes #490

### DIFF
--- a/frankenphp/docker-entrypoint.sh
+++ b/frankenphp/docker-entrypoint.sh
@@ -22,7 +22,7 @@ if [ "$1" = 'frankenphp' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		fi
 	fi
 
-	if [ -z "$(ls -A 'vendor/')" ]; then
+	if [ -z "$(ls -A 'vendor/' 2>/dev/null)" ]; then
 		composer install --prefer-dist --no-progress --no-interaction
 	fi
 
--- a/frankenphp/docker-entrypoint.sh
+++ b/frankenphp/docker-entrypoint.sh
@@ -22,7 +22,7 @@ if [ "$1" = 'frankenphp' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		fi
 	fi
 
-	if [ ! -d 'vendor/' ]; then
+	if [ -z "$(ls -A 'vendor/')" ]; then
 		composer install --prefer-dist --no-progress --no-interaction
 	fi
 


frankenphp/docker-entrypoint.sh
```diff
@@ -22,7 +22,7 @@ if [ "$1" = 'frankenphp' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		fi
 	fi
 
-	if [ ! -d 'vendor/' ]; then
+	if [ -z "$(ls -A 'vendor/')" ]; then
```
dunglas, line 25: Wouldn't the script fail the directory doesn't exist?

bkosun: No, if the `vendor` directory does not exist, the condition will be satisfied

> This condition will be true if the `vendor` directory does not exist (in which case the `ls` command will return an error and the command output will be empty) or if there are no files or subdirectories in the directory (in which case `ls -A` will output empty lines, and the output of the command will also be empty)

Issue #490

vincentchalamon: Tested (locally), the syntax should be:
```suggestion
	if [ -z "$(ls -A 'vendor/' 2>/dev/null)" ]; then
```

vincentchalamon (changes requested):

PR #500, opened 2023-10-30
Include changes for the demo setup

florian25686:


zpi12lmm:
It might be a good idea for a new repository to be named something like this: demo-symfony-docker-frankenphp

PR #517, opened 2023-12-03
feat: add caddy server extra

maxhelias:
tls config should be in the domain config


dunglas:
yes

maxhelias:
@dunglas Done ! ;)

dunglas:
Thank you!!

PR #523, opened 2023-12-06
Add Devcontainer support

ili101:
Copy of https://github.com/api-platform/api-platform/pull/2488 (without the folder change as it's not needed here)


ili101:
Works with VSCode.
I don't use PhpStorm but I see that they added support for Devcontainer but it doesn't work for me, it looks like they are still working on [alpine support](https://youtrack.jetbrains.com/issue/IDEA-333138/DevContainers.-Failed-dev-containers-creation-on-images-without-necessary-dependencies)

maxhelias:
Hello,

Thank you for your contribution.
I am not in favor of adding devcontainer to the project for the following reasons:

- More complexity to the project.
- A very strong opinion in customizations.
- Seems to be scoped only to an IDE (VSCode).
- Is it compatible with extras services support (https://github.com/dunglas/symfony-docker/blob/main/docs/extra-services.md)?

I'm not closed to the idea.. let's talk it over with @dunglas

ili101:
Thank you @maxhelias, some input on the points brought up:
> More complexity to the project.

All changes are under ".devcontainer" and ".vscode", I don't think it adds significant complexity and it's completely optional.
> A very strong opinion in customizations.

This is true.
Some of it can be removed, I added a lot of stuff that is my opinion as an example. I can clean up some of it.
For the rest like for example what PHP extension to use? we can select one by default and comment the rest. or just leave it completely empty for the user.

> Seems to be scoped only to an IDE (VSCode).

Theoretically, it's an open standard https://containers.dev/. Practically works in VSCode and should work in PhpStorm but it's pretty now in PhpStorm so looks like there are some bugs at the moment. As you can see it was reported less than a month ago https://youtrack.jetbrains.com/issue/IDEA-337397/DevContainers.-Creating-fails-for-alpine-images. if this is a blocker we can maybe wait for it to be fixed.

> Is it compatible with extra services support

Should be compatible. also compatible with extra containers setup like in the api-platform template.

maxhelias:
Hello,

With the complexity of maintenance that this integration would add and to the lack of knowledge of this subject, I'd rather not include it in the project.
However, you can have a recipe that brings these configuration files to a project.

Thank you for your proposal.

PR #543, opened 2024-01-11
feat: use install-php-extensions to install Composer

dunglas:


maxhelias (approved):

maxhelias:
Oh interesting! 👍

lenybernard:
Hi,
I don't understand the reason doing this change, could you please explain the point ?

dunglas:
@lenybernard it's mostly a cosmetic change to simplify our Dockerfile. The script we use to install extensions also supports installing Composer, so let's use it instead of installing Composer manually.

Issue #550, opened 2024-01-23
services.php.environment array items[3,12] must be unique


rene-schwabe:
If I run
`docker compose -f compose.yaml -f compose.prod.yaml up -d --wait
`
for production I will get:
`services.php.environment array items[3,12] must be unique`

How do I run the production containers?

CodeCraftDevex:
In compose.prod.yaml 
these variables must be unique, as they are duplicated, causing an error:
 MERCURE_PUBLISHER_JWT_KEY: ${CADDY_MERCURE_JWT_SECRET}
MERCURE_SUBSCRIBER_JWT_KEY: ${CADDY_MERCURE_JWT_SECRET}

PR #551, opened 2024-01-24
Change to unique Variable of composer.prod.yaml #550

rene-schwabe:
Using `docker compose -f compose.yaml -f compose.prod.yaml up -d --wait` causes the following error:

`services.php.environment array items[3,12] must be unique`

Duplicate variables are used in composer.prod.yaml:
MERCURE_PUBLISHER_JWT_KEY: ${CADDY_MERCURE_JWT_SECRET}
MERCURE_SUBSCRIBER_JWT_KEY: ${CADDY_MERCURE_JWT_SECRET}


maxhelias:
I'm not sure this is the right fix. Using an environment variable doesn't cause any problems. I'd say it's the declaration of a docker environment variable that has the problem. Like MERCURE_PUBLISHER_JWT_KEY & MERCURE_SUBSCRIBER_JWT_KEY are configured in compose.yaml and compose.prod.yaml and the up is requested using these 2 files, i guess.

However, I can't reproduce this case. I clone the project, run the command `docker compose -f compose.yaml -f compose.prod.yaml up -d --wait` and I don't have any problems.
Are you working on the latest version of the project? And which docker version are you using?

**Note:** Renaming CADDY_MERCURE_JWT_SECRET also requires adapting all occurrences in the project

PR #567, opened 2024-02-16
Add documentation on how to use MySQL instead of PostgreSQL.

### LINKED ISSUE #566: Extend documention / How-to mysql
I am familiar with MySQL and don't want to use PostgreSQL. I therefore had to make some adjustments for MySQL and there were some pitfalls. Would it make sense to expand the documentation to include this point? If there is interest I can create a PR.

dsentker:
Fixes #566

### DIFF
--- a/docs/mysql.md
+++ b/docs/mysql.md
@@ -50,7 +50,7 @@ Last but not least, we need to install the mysql driver in `Dockerfile`:
 ###< doctrine/doctrine-bundle ###
 ```
 
-## Change environment
+## Change Environment
 ```dotenv 
 DATABASE_URL=mysql://${MYSQL_USER:-app}:${MYSQL_PASSWORD:-!ChangeMe!}@database:3306/${MYSQL_DATABASE:-app}?serverVersion=${MYSQL_VERSION:-8}&charset=${MYSQL_CHARSET:-utf8mb4}
 ```
--- a/docs/mysql.md
+++ b/docs/mysql.md
@@ -8,19 +8,26 @@ First, install the doctrine/orm pack as described: `make composer c='req symfony
 ## Docker configuration
 Change the database image to use MySQL instead of PostgreSQL in `compose.yaml`:
 
-```yaml
+```diff
 ###> doctrine/doctrine-bundle ###
-  database:
-    image: 'mysql:8.0'
+-   image: postgres:${POSTGRES_VERSION:-15}-alpine
++   image: mysql:${MYSQL_VERSION:-8}
     environment:
-      - MYSQL_ROOT_PASSWORD=r007_p4ssw0rd
-      - MYSQL_DATABASE=databasename
-      - MYSQL_USER=mysqluser
-      - MYSQL_PASSWORD=p4ssw0rd
+-     POSTGRES_DB: ${POSTGRES_DB:-app}
++     MYSQL_DATABASE: ${MYSQL_DATABASE:-app}
+      # You should definitely change the password in production
++     MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD:-!ChangeMe!}
+-     POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-!ChangeMe!}
++     MYSQL_PASSWORD: ${MYSQL_PASSWORD:-!ChangeMe!}
+-     POSTGRES_USER: ${POSTGRES_USER:-app}
++     MYSQL_USER: ${MYSQL_USER:-app}
     volumes:
-      - database_data:/var/lib/mysql:rw
+-     - database_data:/var/lib/postgresql/data:rw
++     - database_data:/var/lib/mysql:rw
+      # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!
+-     # - ./docker/db/data:/var/lib/postgresql/data:rw
++     # - ./docker/db/data:/var/lib/mysql:rw
 ###< doctrine/doctrine-bundle ###
-```
 Depending on the database configuration, modify the environment in the same file at `services.php.environment.DATABASE_URL`
 ```
 DATABASE_URL: mysql://mysqluser:p4ssw0rd@database:3306/nucast?serverVersion=8.0.32&charset=utf8mb4
--- a/docs/mysql.md
+++ b/docs/mysql.md
@@ -5,7 +5,7 @@ If you prefer to work with MySQL, some adjustments need to be made.
 
 First, install the doctrine/orm pack as described: `make composer c='req symfony/orm-pack`
 
-## Docker configuration
+## Docker Configuration
 Change the database image to use MySQL instead of PostgreSQL in `compose.yaml`:
 
 ```diff
--- a/docs/mysql.md
+++ b/docs/mysql.md
@@ -42,7 +42,7 @@ Since we changed the port, we also have to define this in the `compose.override.
 ###< doctrine/doctrine-bundle ###
 ```
 
-Last but not least, we need to install the mysql driver in `Dockerfile`:
+Last but not least, we need to install the MySQL driver in `Dockerfile`:
 ```diff
 ###> doctrine/doctrine-bundle ###
 -RUN install-php-extensions pdo_pgsql
--- a/docs/mysql.md
+++ b/docs/mysql.md
@@ -1,7 +1,7 @@
 # Using MySQL
 
-The Docker configuration in this repository defaults to using PostgreSQL.
-If you prefer to work with MySQL, some adjustments need to be made.
+The Docker configuration of this repository is extensible thanks to Flex recipes. By default, the recipe installs PostgreSQL.
+If you prefer to work with MySQL, follow these steps:
 
 First, install the doctrine/orm pack as described: `make composer c='req symfony/orm-pack`
 
--- a/docs/mysql.md
+++ b/docs/mysql.md
@@ -16,7 +16,7 @@ Change the database image to use MySQL instead of PostgreSQL in `compose.yaml`:
 -     POSTGRES_DB: ${POSTGRES_DB:-app}
 +     MYSQL_DATABASE: ${MYSQL_DATABASE:-app}
       # You should definitely change the password in production
-+     MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD:-!ChangeMe!}
++     MYSQL_RANDOM_ROOT_PASSWORD: "true"
 -     POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-!ChangeMe!}
 +     MYSQL_PASSWORD: ${MYSQL_PASSWORD:-!ChangeMe!}
 -     POSTGRES_USER: ${POSTGRES_USER:-app}
--- a/docs/mysql.md
+++ b/docs/mysql.md
@@ -52,7 +52,7 @@ Last but not least, we need to install the mysql driver in `Dockerfile`:
 
 ## Change environment
 ```dotenv 
-DATABASE_URL="mysql://mysqluser:p4ssw0rd@database:3306/databasename?serverVersion=8.0.32&charset=utf8mb4"
+DATABASE_URL=mysql://${MYSQL_USER:-app}:${MYSQL_PASSWORD:-!ChangeMe!}@database:3306/${MYSQL_DATABASE:-app}?serverVersion=${MYSQL_VERSION:-8}&charset=${MYSQL_CHARSET:-utf8mb4}
 ```
 
 ## Final steps
--- a/docs/mysql.md
+++ b/docs/mysql.md
@@ -6,7 +6,7 @@ If you prefer to work with MySQL, some adjustments need to be made.
 First, install the doctrine/orm pack as described: `make composer c='req symfony/orm-pack`
 
 ## Docker configuration
-Change the database image to use mysql instead of postgresql in `compose.yaml`:
+Change the database image to use MySQL instead of PostgreSQL in `compose.yaml`:
 
 ```yaml
 ###> doctrine/doctrine-bundle ###
--- a/docs/mysql.md
+++ b/docs/mysql.md
@@ -1,0 +1,60 @@
+# Using MySQL
+
+The Docker configuration in this repository defaults to using PostgreSQL.
+If you prefer to work with MySQL, some adjustments need to be made.
+
+First, install the doctrine/orm pack as described: `make composer c='req symfony/orm-pack`
+
+## Docker configuration
+Change the database image to use mysql instead of postgresql in `compose.yaml`:
+
+```yaml
+###> doctrine/doctrine-bundle ###
+  database:
+    image: 'mysql:8.0'
+    environment:
+      - MYSQL_ROOT_PASSWORD=r007_p4ssw0rd
+      - MYSQL_DATABASE=databasename
+      - MYSQL_USER=mysqluser
+      - MYSQL_PASSWORD=p4ssw0rd
+    volumes:
+      - database_data:/var/lib/mysql:rw
+###< doctrine/doctrine-bundle ###
+```
+Depending on the database configuration, modify the environment in the same file at `services.php.environment.DATABASE_URL`
+```
+DATABASE_URL: mysql://mysqluser:p4ssw0rd@database:3306/nucast?serverVersion=8.0.32&charset=utf8mb4
+```
+
+Since we changed the port, we also have to define this in the `compose.override.yaml`:
+```yaml
+###> doctrine/doctrine-bundle ###
+  database:
+    ports:
+      - "3306:3306"
+###< doctrine/doctrine-bundle ###
+```
+
+Last but not least, we need to install the mysql driver in `Dockerfile`:
+```diff
+###> doctrine/doctrine-bundle ###
+- RUN install-php-extensions pdo_postgres
++ RUN install-php-extensions pdo_mysql
+###< doctrine/doctrine-bundle ###
+```
+
+## Change environment
+```dotenv 
+DATABASE_URL="mysql://mysqluser:p4ssw0rd@database:3306/databasename?serverVersion=8.0.32&charset=utf8mb4"
+```
+
+## Final steps
+Rebuild the docker environment:
+```shell
+make down && make build
+```
+
+Test your setup:
+```shell
+make sf c='dbal:run-sql -q "SELECT 1" && echo "OK" || echo "Connection is not working"'`
+```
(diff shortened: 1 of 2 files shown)


docs/mysql.md
```diff
@@ -0,0 +1,60 @@
...
+    image: 'mysql:8.0'
+    environment:
+      - MYSQL_ROOT_PASSWORD=r007_p4ssw0rd
+      - MYSQL_DATABASE=databasename
+      - MYSQL_USER=mysqluser
+      - MYSQL_PASSWORD=p4ssw0rd
+    volumes:
+      - database_data:/var/lib/mysql:rw
+###< doctrine/doctrine-bundle ###
+```
```
bkosun, line 23:
```suggestion
```diff
###> doctrine/doctrine-bundle ###
-   image: postgres:${POSTGRES_VERSION:-15}-alpine
+   image: mysql:${MYSQL_VERSION:-8}
    environment:
-     POSTGRES_DB: ${POSTGRES_DB:-app}
+     MYSQL_DATABASE: ${MYSQL_DATABASE:-app}
      # You should definitely change the password in production
+     MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD:-!ChangeMe!}
-     POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-!ChangeMe!}
+     MYSQL_PASSWORD: ${MYSQL_PASSWORD:-!ChangeMe!}
-     POSTGRES_USER: ${POSTGRES_USER:-app}
+     MYSQL_USER: ${MYSQL_USER:-app}
    volumes:
-     - database_data:/var/lib/postgresql/data:rw
+     - database_data:/var/lib/mysql:rw
      # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!
-     # - ./docker/db/data:/var/lib/postgresql/data:rw
+     # - ./docker/db/data:/var/lib/mysql:rw
###< doctrine/doctrine-bundle ###
```
```

File: docs/mysql.md [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+Depending on the database configuration, modify the environment in the same file at `services.php.environment.DATABASE_URL`
+```
+DATABASE_URL: mysql://mysqluser:p4ssw0rd@database:3306/nucast?serverVersion=8.0.32&charset=utf8mb4
```
bkosun, line 26:
```suggestion
DATABASE_URL: mysql://${MYSQL_USER:-app}:${MYSQL_PASSWORD:-!ChangeMe!}@database:3306/${MYSQL_DATABASE:-app}?serverVersion=${MYSQL_VERSION:-8}&charset=${MYSQL_CHARSET:-utf8mb4}
```

docs/mysql.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+Since we changed the port, we also have to define this in the `compose.override.yaml`:
+```yaml
+###> doctrine/doctrine-bundle ###
+  database:
+    ports:
+      - "3306:3306"
+###< doctrine/doctrine-bundle ###
+```
```
bkosun, line 36:
```suggestion
```diff
###> doctrine/doctrine-bundle ###
  database:
    ports:
-     - "5432:5432"
+     - "3306:3306"
###< doctrine/doctrine-bundle ###
```
```

File: docs/mysql.md [hunk 4]
```diff
@@ -0,0 +1,60 @@
...
+
+Last but not least, we need to install the mysql driver in `Dockerfile`:
+```diff
+###> doctrine/doctrine-bundle ###
+- RUN install-php-extensions pdo_postgres
++ RUN install-php-extensions pdo_mysql
```
bkosun, line 42:
```suggestion
-RUN install-php-extensions pdo_pgsql
+RUN install-php-extensions pdo_mysql
```

docs/mysql.md
```diff
@@ -0,0 +1,60 @@
...
+###< doctrine/doctrine-bundle ###
+```
+
+## Change environment
+```dotenv 
+DATABASE_URL="mysql://mysqluser:p4ssw0rd@database:3306/databasename?serverVersion=8.0.32&charset=utf8mb4"
```
bkosun, line 48:
```suggestion
DATABASE_URL=mysql://${MYSQL_USER:-app}:${MYSQL_PASSWORD:-!ChangeMe!}@database:3306/${MYSQL_DATABASE:-app}?serverVersion=${MYSQL_VERSION:-8}&charset=${MYSQL_CHARSET:-utf8mb4}
```

amir-ys:
plz accept this pr
it is user friendly

bkosun (changes requested):

docs/mysql.md
```diff
@@ -0,0 +1,60 @@
+# Using MySQL
+
+The Docker configuration in this repository defaults to using PostgreSQL.
+If you prefer to work with MySQL, some adjustments need to be made.
+
+First, install the doctrine/orm pack as described: `make composer c='req symfony/orm-pack`
```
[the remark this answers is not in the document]
dunglas: We don't use make in the default template, please use compose directly.

docs/mysql.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Docker configuration
```
dunglas, line 8:
```suggestion
## Docker Configuration
```

docs/mysql.md
```diff
@@ -0,0 +1,60 @@
...
+Change the database image to use mysql instead of postgresql in `compose.yaml`:
```
dunglas, line 9:
```suggestion
Change the database image to use MySQL instead of PostgreSQL in `compose.yaml`:
```

docs/mysql.md
```diff
@@ -0,0 +1,60 @@
...
+
+Last but not least, we need to install the mysql driver in `Dockerfile`:
+```diff
+###> doctrine/doctrine-bundle ###
+- RUN install-php-extensions pdo_postgres
++ RUN install-php-extensions pdo_mysql
+###< doctrine/doctrine-bundle ###
+```
+
+## Change environment
```
dunglas, line 46:
```suggestion
## Change Environment
```

dunglas (approved):
LGTM. Should we recommend MySQL or MariaDB?

amir-ys:
> LGTM. Should we recommend MySQL or MariaDB?

Mysql

dunglas:
Why? Maria has more features, is faster and has a larger community.

dsentker:
Thank you for your improvement.

MariaDB hast more Features and better Performance, right. But i'm not sure about Community. Maybe We should consider [popularity](https://db-engines.com/en/ranking_trend/system/MariaDB%3BMySQL)?

docs/mysql.md
dsentker: yes, my intention was to mention the image names. I think that we should use DBMS Name instead.

docs/mysql.md
```diff
@@ -0,0 +1,60 @@
+# Using MySQL
+
+The Docker configuration in this repository defaults to using PostgreSQL.
+If you prefer to work with MySQL, some adjustments need to be made.
```
maxhelias, line 4:
```suggestion
The Docker configuration of this repository is extensible thanks to Flex recipes. By default, the recipe installs PostgreSQL.
If you prefer to work with MySQL, follow these steps:
```
WDYT ?

docs/mysql.md
```diff
@@ -0,0 +1,60 @@
...
+Since we changed the port, we also have to define this in the `compose.override.yaml`:
+```yaml
+###> doctrine/doctrine-bundle ###
+  database:
+    ports:
+      - "3306:3306"
+###< doctrine/doctrine-bundle ###
+```
+
+Last but not least, we need to install the mysql driver in `Dockerfile`:
```
bkosun, line 38:
```suggestion
Last but not least, we need to install the MySQL driver in `Dockerfile`:
```

bkosun (changes requested):

gponty:
Hi @dsentker ,
Thank you for this PR.

But I think there will be a problem with the test database because when mysql/mariadb creates the database and the user for the first time, it only grants privileges on the MYSQL_DATABASE and not the privileges to create a new database.

Therefore, it will not be possible to create the test database unless the user's rights are manually updated beforehand, but the same problem will occur during the continuous integration (CI) execution.

Or maybe I am wrong and there is a solution.

amir-ys:
> Hi @dsentker ,
> Thank you for this PR.
> 
> But I think there will be a problem with the test database because when mysql/mariadb creates the database and the user for the first time, it only grants privileges on the MYSQL_DATABASE and not the privileges to create a new database.
> 
> Therefore, it will not be possible to create the test database :
> 
> ```
> docker compose exec -T php bin/console -e test doctrine:database:create
> An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1044 Access denied for user 'app'@'%' to database 'app_test'
> ```
> 
> unless the user's rights are manually updated beforehand, but the same problem will occur during the continuous integration (CI) execution.
> 
> Or maybe I am wrong and there is a solution.

Actually, I also have this problem

maxhelias:
@gponty @amir-ys Have you used @bkosun's suggestions to unifies database connection ?

gponty:
> @gponty @amir-ys Have you used @bkosun's suggestions to unifies database connection ?

Yes

dsentker:
> But I think there will be a problem with the test database because when mysql/mariadb creates the database and the user for the first time, it only grants privileges on the MYSQL_DATABASE and not the privileges to create a new database.


Oh, that wasn't on my radar. Good catch. In my cases, the database always existed before the container was created.

My guess is that you can only create the database with the `MYSQL_ROOT_PASSWORD`. If I see it correctly, this would have to be done at the time of  `docker compose up`.

But for some reason the environment variables don't work for me at all, so I can only make assumptions here.

Perhaps this point should deliberately be left open and mentioned in the Markdown?

gponty:
> > But I think there will be a problem with the test database because when mysql/mariadb creates the database and the user for the first time, it only grants privileges on the MYSQL_DATABASE and not the privileges to create a new database.
> 
> Oh, that wasn't on my radar. Good catch. In my cases, the database always existed before the container was created.
> 
> My guess is that you can only create the database with the `MYSQL_ROOT_PASSWORD`. If I see it correctly, this would have to be done at the time of `docker compose up`.
> 
> But for some reason the environment variables don't work for me at all, so I can only make assumptions here.
> 
> Perhaps this point should deliberately be left open and mentioned in the Markdown?

Good idea, I will run some tests by logging in as root and I will potentially update your pull request.

docs/mysql.md
```diff
@@ -0,0 +1,70 @@
+# Using MySQL
+
+The Docker configuration of this repository is extensible thanks to Flex recipes. By default, the recipe installs PostgreSQL.
+If you prefer to work with MySQL, follow these steps:
+
+First, install the doctrine/orm pack as described: `docker compose exec php composer req symfony/orm-pack`
```
bkosun, line 6:
```suggestion
First, install the `symfony/orm-pack` package as described: `composer require symfony/orm-pack`
```

docs/mysql.md
```diff
@@ -0,0 +1,70 @@
...
+DATABASE_URL: mysql://${MYSQL_USER:-app}:${MYSQL_PASSWORD:-!ChangeMe!}@database:3306/${MYSQL_DATABASE:-app}?serverVersion=${MYSQL_VERSION:-8}&charset=${MYSQL_CHARSET:-utf8mb4}
+```
+
+Since we changed the port, we also have to define this in the `compose.override.yaml`:
+```diff
+###> doctrine/doctrine-bundle ###
+  database:
+    ports:
+-     - "5432:5432"
++     - "3306:3306"
```
bkosun, line 44:
```suggestion
-     - "5432"
+     - "3306"
```

bkosun (changes requested):

docs/mysql.md
bkosun: Installing the `symfony/orm-pack` package should not be done inside a container

maxhelias: Normally, it should be inside the container, but in our case it can be a bit tricky.
These files are listed in the `.dockerignore` file but shared in the `compose.override.yaml` and at this stage we are more in dev mode where these files are shared.

dsentker: Maybe I'm misunderstanding you, but isn't it exactly right? Working with composer outside the container is error-prone because the local PHP version does not necessarily match or other parameters are different (extensions, etc.)

bkosun:
@gponty @amir-ys, @dsentker

Once the container is built, the database will be created automatically based on the configuration

```shell
$ docker compose exec php bin/console doctrine:database:create
Could not create database `app` for connection named default
An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1007 Can't create database 'app'; database exists
```

docs/mysql.md
bkosun: The Symfony recipe applies to `compose.yml`, `compose.override.yaml`, `Dockerfile`, etc.

If the container is running in development mode, the changes will be applied in the local folder thanks to this line:
https://github.com/dunglas/symfony-docker/blob/f063b11fb8d3522e3811124f43b2145dad205199/compose.override.yaml#L8

Otherwise (production mode) it won't work, so I would recommend using this command outside the container to avoid any confusion. Anyway, this is just a recommendation. Perhaps we should use some uniform style for installing additional packages.

bkosun: The Symfony recipe applies to `compose.yml`, `compose.override.yaml`, `Dockerfile`, etc.

If the container is running in development mode, the changes will be applied in the local folder:
https://github.com/dunglas/symfony-docker/blob/f063b11fb8d3522e3811124f43b2145dad205199/compose.override.yaml#L8

Otherwise (production mode) it won't work, so I would recommend using this command outside the container to avoid any confusion. Anyway, this is just a recommendation. Perhaps we should use some uniform style for installing additional packages.

dsentker:
> @gponty @amir-ys, @dsentker
> 
> Once the container is built, the database will be created automatically based on the configuration
> 
> ```shell
> $ docker compose exec php bin/console doctrine:database:create
> Could not create database `app` for connection named default
> An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1007 Can't create database 'app'; database exists
> ```

So we're good to go & no more documentation required?

docs/mysql.md
maxhelias: In production mode, it won't work and I'm not sure we'll end up in that this situation, so i'm in favour of doing it inside.
However, it deserves to be mentioned see : https://github.com/dunglas/symfony-docker/pull/571/files

maxhelias: You're right @dsentker. The point raised by @bkosun is the files modified by recipes are not sync in the container in prod mode. So the the recipe can't be applied.

amir-ys:
> > @gponty @amir-ys, @dsentker
> > 
> > Once the container is built, the database will be created automatically based on the configuration
> > 
> > ```shell
> > $ docker compose exec php bin/console doctrine:database:create
> > Could not create database `app` for connection named default
> > An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1007 Can't create database 'app'; database exists
> > ```
> 
> So we're good to go & no more documentation required?

I will test soon and announce the result

docs/mysql.md
dsentker: > However, you raise a good point that deserves to be mentioned see : https://github.com/dunglas/symfony-docker/pull/571/files

Maybe add a second sentence like 

> [!NOTE] Be sure that your host machine does provide  the same requirements as defined in the composer configuration files.

gponty:
> > > @gponty @amir-ys, @dsentker
> > > Once the container is built, the database will be created automatically based on the configuration
> > > ```shell
> > > $ docker compose exec php bin/console doctrine:database:create
> > > Could not create database `app` for connection named default
> > > An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1007 Can't create database 'app'; database exists
> > > ```
> > 
> > 
> > So we're good to go & no more documentation required?
> 
> I will test soon and announce the result

The only way I found is to modify the DATABASE_URL to put root user; after that, I don't know if there's an impact on production (I imagine there is, but I only use this image in development), but env dev and CI is ok.

`
DATABASE_URL: mysql://root:${MYSQL_ROOT_PASSWORD:-!ChangeMeComposer!}@database:3306/${MYSQL_DATABASE:-app}?serverVersion=${MYSQL_VERSION:-8}&charset=${MYSQL_CHARSET:-utf8mb4}`

7-zete-7:
@gponty, do you clear the storage (linked to `/var/lib/mysql`) before launching with changed parameters?

The documentation for `mysql` image [says](https://hub.docker.com/_/mysql#:~:text=If%20a%20user/password%20was%20supplied%20(see%20below)%20then%20that%20user%20will%20be%20granted%20superuser%20access%20(corresponding%20to%20GRANT%20ALL)%20to%20this%20database.) that configured users will have granted superuser access to the configured database.

The container probably starts with the database already initialized.

gponty:
> @gponty, do you clear the storage (linked to `/var/lib/mysql`) before launching with changed parameters?
> 
> The documentation for `mysql` image [says](https://hub.docker.com/_/mysql#:~:text=If%20a%20user/password%20was%20supplied%20(see%20below)%20then%20that%20user%20will%20be%20granted%20superuser%20access%20(corresponding%20to%20GRANT%20ALL)%20to%20this%20database.) that configured users will have granted superuser access to the configured database.
> 
> The container probably starts with the database already initialized.

No need, just restart the container so it connects with user root, and when you create your test database, it will be able to create it because the user will be root.

docs/mysql.md
```diff
@@ -0,0 +1,70 @@
...
+
+```diff
+###> doctrine/doctrine-bundle ###
+-   image: postgres:${POSTGRES_VERSION:-15}-alpine
++   image: mysql:${MYSQL_VERSION:-8}
+    environment:
+-     POSTGRES_DB: ${POSTGRES_DB:-app}
++     MYSQL_DATABASE: ${MYSQL_DATABASE:-app}
+      # You should definitely change the password in production
++     MYSQL_ROOT_PASSWORD: ${MYSQL_ROOT_PASSWORD:-!ChangeMe!}
```
maxhelias, line 19:
```suggestion
+     MYSQL_RANDOM_ROOT_PASSWORD: "true"
```

docs/mysql.md
```diff
@@ -0,0 +1,70 @@
...
+
+Last but not least, we need to install the MySQL driver in `Dockerfile`:
+```diff
+###> doctrine/doctrine-bundle ###
+-RUN install-php-extensions pdo_pgsql
++RUN install-php-extensions pdo_mysql
+###< doctrine/doctrine-bundle ###
+```
+
+## Change Environment
```
maxhelias, line 56:
```suggestion
## Change Environment
Change the database configuration in `.env`:

```

maxhelias:
@dsentker I've just applied the last 3 suggestions, tested the procedure and everything works perfectly! Thanks!

maxhelias:
Thanks also @bkosun for your review

PR #571, opened 2024-02-19
docs: add mention to use composer require inside container

maxhelias:


dsentker (approved):

develth:
hey, 

IMHO should there be a info, that if you have `- /app/vendor` within `compose.override.yaml` that it will not sync to your local files, if your run it wihin container.

maxhelias:
Hi!
Even if the `vendor` folder is no longer synchronized, this is not a problem for configuring extras service via recipes, since it doesn't modify the vendor but the root files, which are synchronized.

develth:
Hi,
yes, but if you have for example an IDE that checks imports etc, there is an error, because the files that are referenced in a bundle are not resolved / known.
If you are not aware of thats why vendor is not synced to your local folder, you will have a long search for errors (speaking with experience :D)

maxhelias:
But this has nothing to do with extra service.

develth:
Just thought a reminder would be useful for people reading doc & maybe wonder, why they can´t find the files. 
Nevermind.

Issue #537, opened 2024-01-03
Debian version of Dockerfile


rccc:
Hi, I'm wondering if there is a debian version of the Dockerfile (I've read that there is a debian version for frankenphp).

Thanks in advance

maxhelias:
Hi, we don't have a debian version here.
We try to have a template that is as optimized as possible for production, development and CI. Covering different approaches is not our goal.
Thanks

PR #582, opened 2024-03-07
feat: switch to Debian instead of Alpine

### LINKED ISSUE #555: Moving to Debian
There are some weird issues with Alpine and PHP, and I recently learned that musl isn't really actively supported by PHP. They fix bugs when reported or patched, but don't run tests with musl etc. There are also many known issues with JIT and musl.

I propose to switch back to Debian/glibc, which is officially supported by PHP.

dunglas:
Closes #555, #537.

Maybe could we even explain how to switch to Alpine in a documentation entry? It's super easy.

### DIFF
--- a/frankenphp/conf.d/app.ini
+++ b/frankenphp/conf.d/app.ini
@@ -1,4 +1,3 @@
-variables_order = EGPCS
 expose_php = 0
 date.timezone = UTC
 apc.enable_cli = 1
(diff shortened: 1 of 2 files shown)


frankenphp/conf.d/app.ini
```diff
@@ -1,4 +1,3 @@
-variables_order = EGPCS
```
dunglas, line 1: This is unrelated and should be done on Alpine too.

PR #584, opened 2024-03-12
docs: using Alpine Linux instead of Debian

dunglas:
See #582.


maxhelias (approved):

maxhelias:
Thanks for documenting this

Issue #538, opened 2024-01-05
php config file order


wizhippo:
As php takes the last ini value set, in frankenphp/conf.d

The order currently is:

app.dev.ini
app.ini

Expected would be:

app.ini
app.dev.ini

Should app.dev.ini be renamed so it is loaded after app.ini?  Currently there is only one setting in app.dev.ini but more could be added and I believe the expected behavior would be it take precedence over settings in app.ini.

maxhelias:
Yes, a PR would be welcome.
We currently have these : 
<img width="571" alt="Capture d’écran 2024-02-25 à 19 36 37" src="https://github.com/dunglas/symfony-docker/assets/12966574/8cbd2986-72c4-41e8-9b7d-eb2762cc5760">

wizhippo:
Is there any preference?

rename it to app.zzdev.ini or zzz-app.dev.ini?

maxhelias:
With number ? `00-app.ini` & `01-app.dev.ini` ?

wizhippo:
> With number ? `00-app.ini` & `01-app.dev.ini` ?

Numbers will move them above the docker-php-ext-xyz.ini  I would think we want our config to supersede them so I used the z prefix to move them to the bottom as php uses the last set value.  So maybe z01-app.ini, z02-other.ini, ... Again normally the dev is loaded last so zzz-dev.ini seems most likely to be loaded last.

leroy0211:
You could also set the `PHP_INI_SCAN_DIR` environment variable to 2 paths. 

```
ENV PHP_INI_SCAN_DIR="$PHP_INI_DIR/conf.d:$PHP_INI_DIR/app.conf.d"
```

This way the `conf.d` directory is scanned first for ini files, and then the `app.conf.d` is scanned for ini files.  

This way you can sort the app's ini files in numerical order. 

```
/usr/local/etc/php/app.conf.d/01-app.ini
/usr/local/etc/php/app.conf.d/09-app.dev.ini  - or -  /usr/local/etc/php/app.conf.d/09-app.prod.ini
```

This way you can leave some room for developers to add ini config in all locations.

PR #593, opened 2024-03-21
fix: php ini loading order

wizhippo:
https://github.com/dunglas/symfony-docker/issues/538

As php takes the last ini value set, in frankenphp/conf.d

The order currently is:

app.dev.ini
app.ini

Expected would be:

app.ini
app.dev.ini

This renames app.dev.ini to app.zzz-dev.ini so it is loaded last.


wizhippo:
Perhaps need to be changed again so both app.ini and app.dev.ini are both loaded after all others?

app.ini ->  zzz-app.ini
add.dev.ini -> zzz-dev.ini

dunglas:
Maybe 01- and 02- ?

wizhippo:
> #538

I commented in the issue but wont leading with numbers load before any docker-php-ext-xyz.ini configs?  Normally do we not want app to supersede those too?  That's why I choose the zzz prefix.  Also in the case of zzz-dev.ini, that seems more likely to be always last. even if adding other user config files.

maxhelias:
Yes, you're right. So we need something that passes after `docker-php-ext`.
What do you think about renaming `app` in `php` in this case for files to come after extensions and then choosing how to put the dev last?

leroy0211:
I've opened another PR #617 based on my comment in #538.

PR #594, opened 2024-03-27
Added Git specific instuctions for copying the files to an existing project

m1mir:
With the new instructions the copying to an existing project doesn't destroy the cloned symfony-docker repository.

### DIFF
--- a/docs/existing-project.md
+++ b/docs/existing-project.md
@@ -2,9 +2,13 @@
 
 It's also possible to use Symfony Docker with existing projects!
 
-First, [download this skeleton](https://github.com/dunglas/symfony-docker). If you clone the Git repository, be sure to remove the `.git` directory to prevent conflicts with the `.git` directory already in your existing project.
+First, [download this skeleton](https://github.com/dunglas/symfony-docker).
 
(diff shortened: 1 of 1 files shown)


docs/existing-project.md
```diff
@@ -2,9 +2,13 @@
 
 It's also possible to use Symfony Docker with existing projects!
 
-First, [download this skeleton](https://github.com/dunglas/symfony-docker). If you clone the Git repository, be sure to remove the `.git` directory to prevent conflicts with the `.git` directory already in your existing project.
+First, [download this skeleton](https://github.com/dunglas/symfony-docker).
 
-Then, copy the Docker-related files from the skeleton to your existing project:
+If you cloned the Git repository you can copy the files from the repository using git and tar:
+
+    git archive --format=tar HEAD | tar -xC my-existing-project/
```
maxhelias, line 9: That's good tip, but I think we should always explain why it's necessary to remove the `.git` directory to avoid conflicts

m1mir: Re added the warning with slight modification.

maxhelias: Perfect! Thanks @m1mir

Issue #602, opened 2024-04-09
Should var/ be a volume in production container based on compose or Dockerfile


riskersen:
I noticed, that overlay2 folder of a recently+new deployed is increasing a lot in size. Therefore, I checked with docker diff what's causing the extension of the size and most of the files are related to /app/var or /tmp

I already checked some of the issues here, e.g. #489 and there it's said, that /app/var is considered as VOLUME in the Dockerfile in production, but forgive me, if I'm wrong, the VOLUME part only applies to frankenphp_dev and not to prod build target, so I'm wondering if that's missing in the Dockerfile or in the sample compose file?

As far as I understood, var/ is changing constantly, so it should be a volume as well, no?

Thanks!

maxhelias:
Hello,

Yes, the VOLUME instruction should surely be part of the image base in the `Dockerfile`

PR #608, opened 2024-04-18
fix: volume instruction moved for all stages

riskersen:
…ixes #602

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,9 @@ FROM frankenphp_upstream AS frankenphp_base
 
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -13,6 +13,9 @@ FROM frankenphp_upstream AS frankenphp_base
 
 WORKDIR /app
 
+# symfony loves caching but to keep the overlay clean, var/ shall be a volume on all stages
+VOLUME /app/var/
```
maxhelias, line 17: No need to comment it
```suggestion
VOLUME /app/var/
```

PR #617, opened 2024-04-25
Fix to load app's ini files after all other extensions are loaded

### LINKED ISSUE #538: php config file order

leroy0211:
Fixes #538 by specifying a different `conf.d` directory (`app.conf.d`) for loading the ini files. 

Renamed the files to numerical prefixes so they are loaded in the correct order. Default first, then environment specific. 

After running a dev build, this is the output of `php --ini`

```
Configuration File (php.ini) Path: /usr/local/etc/php
Loaded Configuration File:         /usr/local/etc/php/php.ini
Scan for additional .ini files in: /usr/local/etc/php/conf.d:/usr/local/etc/php/app.conf.d
Additional .ini files parsed:      /usr/local/etc/php/conf.d/docker-php-ext-apcu.ini,
/usr/local/etc/php/conf.d/docker-php-ext-intl.ini,
/usr/local/etc/php/conf.d/docker-php-ext-opcache.ini,
/usr/local/etc/php/conf.d/docker-php-ext-sodium.ini,
/usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini,
/usr/local/etc/php/conf.d/docker-php-ext-zip.ini,
/usr/local/etc/php/app.conf.d/10-app.ini,
/usr/local/etc/php/app.conf.d/20-app.dev.ini
```


frankenphp/conf.d/10-app.ini
leroy0211, line 1: I haven't worked with github templates before, so I'm not sure if this needs to be backwards compatible or not.

Dockerfile
```diff
@@ -36,10 +36,12 @@ RUN set -eux; \
 # https://getcomposer.org/doc/03-cli.md#composer-allow-superuser
 ENV COMPOSER_ALLOW_SUPERUSER=1
 
+ENV PHP_INI_SCAN_DIR="$PHP_INI_DIR/conf.d:$PHP_INI_DIR/app.conf.d"
```
bkosun, line 39:
```suggestion
ENV PHP_INI_SCAN_DIR=":$PHP_INI_DIR/app.conf.d"
```

bkosun (changes requested):

Dockerfile
bkosun: In this case, PHP is configured with the `--with-config-file-scan-dir=/usr/local/etc/php/conf.d` option, so there is no need to specify this directory manually

leroy0211: I tried that, but that didn't worked.  Seems to be working now. I'll apply it

frankenphp/conf.d/10-app.ini
dunglas: No need for backward compatibility!

leroy0211: Great!

goulashsoup:
@dunglas I there a reason this is not merged yet?

dunglas (approved):
LGTM. WDYT @maxhelias

maxhelias (approved):
I agree with this 👍

maxhelias:
Thanks @leroy0211 for fixing this and @bkosun for the review 😉

PR #625, opened 2024-04-30
fix starting container process caused: exec: "bin/phpunit": permission

EnterVPL:
Hi,
While checking the CI template, I encountered an error: `OCI runtime exec failed: exec failed: container_linux.go:370: starting container process caused: exec: "bin/phpunit": permission.` It surprised me a lot because everything was working fine locally. I managed to find a solution to the problem here: https://github.com/api-platform/api-platform/issues/1819. I added a patch to the docker-entrypoint.sh file:

```shell
if [ "$APP_ENV" != 'prod' ]; then
   # https://github.com/api-platform/api-platform/issues/1819
   chmod +x bin/phpunit
fi
```
and everything started functioning correctly.

### DIFF
--- a/frankenphp/docker-entrypoint.sh
+++ b/frankenphp/docker-entrypoint.sh
@@ -51,6 +51,10 @@ if [ "$1" = 'frankenphp' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		if [ "$( find ./migrations -iname '*.php' -print -quit )" ]; then
 			php bin/console doctrine:migrations:migrate --no-interaction --all-or-nothing
 		fi
+		if [ "$APP_ENV" != 'prod' ]; then
+   			# https://github.com/api-platform/api-platform/issues/1819
+   			chmod +x bin/phpunit
+		fi
 	fi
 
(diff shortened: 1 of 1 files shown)


7-zete-7:
Hello @EnterVPL!

It will be enough to commit the `bin/phpunit` file with the rights `0775`. In this case, this file will be executable in the repository and there will be no need to perform this fix.

EnterVPL:
Hello

@7-zete-7 and @maxhelias 

I tried using the `chmod +x bin/phpunit` command via docker and git bash, but it didn't help.
The current file permissions are
```bash
$ ls -la | grep phpunit
-rwxr-xr-x 1 user 197121 776 May  1 17:43 phpunit*
```
I use Windows and I am one of many who need a universal solution to the problem of automatically running tests.
There are no errors in the local environment. They are only on github actions

7-zete-7:
@EnterVPL, when using Windows, I can recommend switching to store the project in WSL2 to solve a number of problems.

When working directly in Windows, you need to use a special git command and options (https://stackoverflow.com/a/40979016) to set the file's permissions.

PR #640, opened 2024-06-16
Update the documentation to change from PostgreSQL to MySQL.

hammat:
The install of doctrine-bundle adds a healthcheck section that was not documented.
Of course, GitHub CI was failing with an unhealthy status because of the PostgreSQL command. 

Update the documentation on how to use MySQL.


dunglas (approved):

maxhelias (approved):

maxhelias:
Thanks @hammat !

PR #651, opened 2024-08-07
chore: remove useless trusted proxies environement variables

dunglas:
See https://github.com/symfony/symfony/pull/37357 and https://github.com/symfony/recipes/pull/790

### DIFF
--- a/compose.override.yaml
+++ b/compose.override.yaml
@@ -6,6 +6,7 @@ services:
       target: frankenphp_dev
     volumes:
       - ./:/app
+      - /app/var
       - ./frankenphp/Caddyfile:/etc/caddy/Caddyfile:ro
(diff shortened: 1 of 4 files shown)


compose.override.yaml
```diff
@@ -6,6 +6,7 @@ services:
       target: frankenphp_dev
     volumes:
       - ./:/app
+      - /app/var
```
maxhelias, line 9: Are you sure for this ? Or should we put it at the same level as the `#- app/vendor` line just below ?

maxhelias: What I mean is that I personally treat `var` & `vendor` in the same way on my projects.

dunglas: Sorry, this was a leftover of an unrelated experiment. Fixed.

maxhelias (approved):

PR #652, opened 2024-08-08
Fix Docker development volumes location for the 20-app.dev.ini file

Trim:
With the fix to load app's ini files after all other extensions are loaded (#617), the app-dev.ini file has been renamed 20-app.dev.ini and has been moved to a folder named "app.conf.d".

Although, the fix forgot to update the development environment file `compose.override.yaml` with the new name and location.

### DIFF
--- a/compose.override.yaml
+++ b/compose.override.yaml
@@ -7,7 +7,7 @@ services:
     volumes:
       - ./:/app
       - ./frankenphp/Caddyfile:/etc/caddy/Caddyfile:ro
-      - ./frankenphp/conf.d/app.dev.ini:/usr/local/etc/php/conf.d/app.dev.ini:ro
+      - ./frankenphp/conf.d/20-app.dev.ini:/usr/local/etc/php/app.conf.d/20-app.dev.ini:ro
       # If you develop on Mac or Windows you can remove the vendor/ directory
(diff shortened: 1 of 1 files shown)


Trim:
Without this fix, docker raises this error:

```
$ docker compose up
[+] Running 1/0
 ✔ Container symfony-docker-php-1  Recreated                                                                                                                                                                                                                                                                                                                                                                                          0.1s
Attaching to php-1
Gracefully stopping... (press Ctrl+C again to force)
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/tmp/symfony-docker/frankenphp/conf.d/app.dev.ini" to rootfs at "/usr/local/etc/php/conf.d/app.dev.ini": mount /tmp/symfony-docker/frankenphp/conf.d/app.dev.ini:/usr/local/etc/php/conf.d/app.dev.ini (via /proc/self/fd/7), flags: 0x5000: not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
```

dunglas (approved):

dunglas:
Thanks!

PR #662, opened 2024-09-05
Install libnss3-tools so Caddy can use certutil for a root certificate

Gramzivi:
Caddy is trying to install a root certificate but can't find certutil. This would help integrate certificates into the local trust store, which may be useful during local development.

The issue can be seen during the starting container:

     INFO    warning: "certutil" is not available, install "certutil" with "apt install libnss3-tools" or "yum install nss-tools" and try again

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 	file \
 	gettext \
 	git \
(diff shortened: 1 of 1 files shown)


dunglas:
This is just a warning that can be safely ignored in most cases: https://github.com/caddyserver/caddy-docker/issues/303

PR #663, opened 2024-09-07
chore: do not specify minor version of Dockerfile syntax

pan93412:
> We recommend using `docker/dockerfile:1`, which always points to the latest stable release of the version 1 syntax, and receives both "minor" and "patch" updates for the version 1 release cycle. 

<https://docs.docker.com/build/buildkit/frontend/>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-#syntax=docker/dockerfile:1.4
+#syntax=docker/dockerfile:1
 
 # Versions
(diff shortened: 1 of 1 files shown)


dunglas:
Thanks! Cool you also propose the patch on https://github.com/api-platform/api-platform, please?

pan93412:
> Thanks! Could you also propose the patch on https://github.com/api-platform/api-platform, please?

Sure! I will do it later ;)

dunglas:
Thanks!

PR #671, opened 2024-09-24
Remove symfony.lock copy in Dockerfile

Zayon:
As composer install in run with `--no-scripts` option, symfony flex will not run so symfony.lock is not required

**This PR is more of a question than an actual PR**


maxhelias:
The symfony.lock file can still be used for consulting recipes. The `composer recipes` command depends on it, for example.
I think it should be kept.

Issue #539, opened 2024-01-09
Symfony Messenger


norkunas:
How to run Symfony Messenger with this setup?

I thought about this in compose.yaml:
```yaml
  php-worker:
    build:
      context: .
      target: frankenphp_dev
    image: ${IMAGES_PREFIX:-}app-php
    restart: unless-stopped
    command: php bin/console messenger:consume async -v
```

Does it make sense? What are the options? Would be good to add something about this to README :)

dunglas:
I use something similar in my projects.

This definitely is worth a documentation section!

MrSpoocy:
Same here, I'm not sure if I create a second container or maybe frankenphp can handle the consumer in a second thread.

norkunas:
noticed one thing with this approach, that all the images at startup tries to run database migrations

MichaelBrauner:
I think that's totally fine. 
Every container will try to run migrations. 
As long as you have a single source of truth for your migrations (one database server for all containers) you should be fine.

norkunas:
Yes, the first one to run wins :) but I am thinking for a flag that I'd set on the main container, and on messenger/scheduler containers not setting it then entrypoint could run based on that condition, is that possible?

MichaelBrauner:
Yes, that is of course possible: 

```yaml
  php-worker:
    build:
      context: .
      target: frankenphp_dev
    image: ${IMAGES_PREFIX:-}app-php
    restart: unless-stopped
    command: php bin/console messenger:consume async -v
    environment:
       - RUN_MIGRATIONS=false
```
And then react to it inside your entry point script:

`if [ "$RUN_MIGRATIONS" = "true" ]; then`

But I personally would advise against it. 
At the moment you just have a simple php container that is no different from the others, except for the command.
That's simple and easy.

Since migrations cannot be executed twice anyway there is no need for additional logic imho. 
But it's up to you.

norkunas:
Thanks :)

norkunas:
> Since migrations cannot be executed twice anyway there is no need for additional logic imho.

Yes, but it's slows down a little bit container startup time :)

MichaelBrauner:
Hi tdumalin,

You might consider using the `--limit` option with `messenger:consume`:

```yaml
php-worker:
  build:
    context: .
    target: frankenphp_dev
  image: ${IMAGES_PREFIX:-}app-php
  restart: unless-stopped
  command: php bin/console messenger:consume async -v --limit=10000
```
[...]

PR #681, opened 2024-10-27
Async Symfony Messenger

k-37:
This is proposal to fix issue #539 based on my interpretation of conversation in the issue.

### DIFF
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -3,15 +3,14 @@
 Add new service to the `compose.yaml`:
 ```yaml
   php-worker:
-  image: ${IMAGES_PREFIX:-}app-php-worker
-  restart: unless-stopped
-  environment:
-    - RUN_MIGRATIONS=false
-  healthcheck:
-    disable: true
-  depends_on:
-    - php
-```
+    image: ${IMAGES_PREFIX:-}app-php-worker
+    restart: unless-stopped
+    environment:
+      - RUN_MIGRATIONS=false
+    healthcheck:
+      disable: true
+    depends_on:
+      - php
 
 Add new services to the `compose.override.yaml`:
 ```yaml
@@ -44,18 +43,16 @@ To add additional workers just copy `php-worker-async` service and replace every
 Add new service to the `compose.prod.yaml`:
 ```yaml
   php-worker:
-  build:
-    context: .
-    target: frankenphp_prod
-  environment:
-    APP_SECRET: ${APP_SECRET}
-```
+    build:
+      context: .
+      target: frankenphp_prod
+    environment:
+      APP_SECRET: ${APP_SECRET}
 
 Apply the following changes to the `frankenphp/docker-entrypoint.sh`:
 ```patch
 -	if grep -q ^DATABASE_URL= .env; then
-+	run_migrations=${RUN_MIGRATIONS:-true}
-+	if grep -q ^DATABASE_URL= .env && [ "$run_migrations" = "true" ]; then
++	if grep -q ^DATABASE_URL= .env && [ "${RUN_MIGRATIONS:-true}" = "true" ]; then
 ```
 
 > [!NOTE]
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -52,6 +52,7 @@ Add new services to the `compose.prod.yaml`:
     extends:
       file: compose.yaml
       service: php-worker
+    image: ${IMAGES_PREFIX:-}app-php-worker-async
     build:
       context: .
       target: frankenphp_prod
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -37,9 +37,11 @@ Add new services to the `compose.override.yaml`:
         condition: service_healthy
 ```
 
-Two instances of `php-worker-async` will start after `php` container which does installation of Symfony. They will share app folder because of `- ./:/app` in volumes configuration. `- /app/var/` defines that every container will have its own and separate `/app/var/` folder, [note missing `:`](https://stackoverflow.com/questions/46166304/docker-compose-volumes-without-colon).
+Two instances, which are defined by `scale` property, of `php-worker-async` will start after `php` container which does installation of Symfony and runs database migrations. They will share app folder because of `- ./:/app` in volumes configuration. `- /app/var/` defines that every container will have its own and separate `/app/var/` folder, [note missing `:`](https://stackoverflow.com/questions/46166304/docker-compose-volumes-without-colon).
 
 To add additional workers just copy `php-worker-async` service and replace every usage of the `async` in the new service with appropriate value for the new worker.
+
+In `compose.yaml` we have base `php-worker` service which is extended by `php-worker-async` in `compose.override.yaml` and in `compose.prod.yaml` with `extends` property. [`donotstart` service profile](https://docs.docker.com/compose/how-tos/profiles/) on `php-worker` service is used to prevent it from starting, because it is only used as a base for eventual multiple workers with different configuration.
 
 Add new services to the `compose.prod.yaml`:
 ```yaml
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -52,7 +52,7 @@ Add new services to the `compose.prod.yaml`:
     extends:
       file: compose.yaml
       service: php-worker
-    image: ${IMAGES_PREFIX:-}app-php-worker-async
+    image: php-worker-async
     build:
       context: .
       target: frankenphp_prod
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -11,6 +11,7 @@ Add new service to the `compose.yaml`:
       disable: true
     depends_on:
       - php
+```
 
 Add new services to the `compose.override.yaml`:
 ```yaml
@@ -48,6 +49,7 @@ Add new service to the `compose.prod.yaml`:
       target: frankenphp_prod
     environment:
       APP_SECRET: ${APP_SECRET}
+```
 
 Apply the following changes to the `frankenphp/docker-entrypoint.sh`:
 ```patch
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -28,7 +28,7 @@ Add new services to the `compose.override.yaml`:
     build:
       context: .
       target: frankenphp_dev
-    command: ['/app/bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
+    command: ['bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
     volumes:
       - ./:/app
       - /app/var/
@@ -56,7 +56,7 @@ Add new services to the `compose.prod.yaml`:
     build:
       context: .
       target: frankenphp_prod
-    command: ['/app/bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
+    command: ['bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
     depends_on:
       php:
         condition: service_healthy
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -55,7 +55,7 @@ Add new services to the `compose.prod.yaml`:
     image: ${IMAGES_PREFIX:-}app-php-worker-async
     build:
       context: .
-      target: frankenphp_dev
+      target: frankenphp_prod
     command: ['/app/bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
     depends_on:
       php:
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -52,7 +52,6 @@ Add new services to the `compose.prod.yaml`:
     extends:
       file: compose.yaml
       service: php-worker
-    image: php-worker-async
     build:
       context: .
       target: frankenphp_prod
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -1,0 +1,62 @@
+# Using async Symfony Messenger
+
+Add new service to the `compose.yaml`:
+```yaml
+  php-worker:
+  image: ${IMAGES_PREFIX:-}app-php-worker
+  restart: unless-stopped
+  environment:
+    - RUN_MIGRATIONS=false
+  healthcheck:
+    disable: true
+  depends_on:
+    - php
+```
+
+Add new services to the `compose.override.yaml`:
+```yaml
+  php-worker:
+    profiles:
+      - donotstart
+
+  php-worker-async:
+    scale: 2
+    extends:
+      file: compose.yaml
+      service: php-worker
+    image: ${IMAGES_PREFIX:-}app-php-worker-async
+    build:
+      context: .
+      target: frankenphp_dev
+    command: ['/app/bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
+    volumes:
+      - ./:/app
+      - /app/var/
+    depends_on:
+      php:
+        condition: service_healthy
+```
+
+Two instances of `php-worker-async` will start after `php` container which does installation of Symfony. They will share app folder because of `- ./:/app` in volumes configuration. `- /app/var/` defines that every container will have its own and separate `/app/var/` folder, [note missing `:`](https://stackoverflow.com/questions/46166304/docker-compose-volumes-without-colon).
+
+To add additional workers just copy `php-worker-async` service and replace every usage of the `async` in the new service with appropriate value for the new worker.
+
+Add new service to the `compose.prod.yaml`:
+```yaml
+  php-worker:
+  build:
+    context: .
+    target: frankenphp_prod
+  environment:
+    APP_SECRET: ${APP_SECRET}
+```
+
+Apply the following changes to the `frankenphp/docker-entrypoint.sh`:
+```patch
+-	if grep -q ^DATABASE_URL= .env; then
++	run_migrations=${RUN_MIGRATIONS:-true}
++	if grep -q ^DATABASE_URL= .env && [ "$run_migrations" = "true" ]; then
+```
+
+> [!NOTE]
+> After all changes are made the containers need to be rebuilt.
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -41,7 +41,7 @@ Two instances of `php-worker-async` will start after `php` container which does 
 
 To add additional workers just copy `php-worker-async` service and replace every usage of the `async` in the new service with appropriate value for the new worker.
 
-Add new service to the `compose.prod.yaml`:
+Add new services to the `compose.prod.yaml`:
 ```yaml
   php-worker:
     profiles:
--- a/docs/symfony-messenger.md
+++ b/docs/symfony-messenger.md
@@ -44,11 +44,22 @@ To add additional workers just copy `php-worker-async` service and replace every
 Add new service to the `compose.prod.yaml`:
 ```yaml
   php-worker:
+    profiles:
+      - donotstart
+
+  php-worker-async:
+    scale: 2
+    extends:
+      file: compose.yaml
+      service: php-worker
+    image: ${IMAGES_PREFIX:-}app-php-worker-async
     build:
       context: .
-      target: frankenphp_prod
-    environment:
-      APP_SECRET: ${APP_SECRET}
+      target: frankenphp_dev
+    command: ['/app/bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
+    depends_on:
+      php:
+        condition: service_healthy
 ```
 
 Apply the following changes to the `frankenphp/docker-entrypoint.sh`:
--- a/README.md
+++ b/README.md
@@ -39,8 +39,9 @@ with [FrankenPHP](https://frankenphp.dev) and [Caddy](https://caddyserver.com/) 
 7. [Using MySQL instead of PostgreSQL](docs/mysql.md)
 8. [Using Alpine Linux instead of Debian](docs/alpine.md)
 9. [Using a Makefile](docs/makefile.md)
-10. [Updating the template](docs/updating.md)
-11. [Troubleshooting](docs/troubleshooting.md)
+10. [Using async Symfony Messenger](docs/symfony-messenger.md)
+11. [Updating the template](docs/updating.md)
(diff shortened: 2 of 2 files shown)


docs/symfony-messenger.md
```diff
@@ -0,0 +1,62 @@
...
+  php-worker:
+  image: ${IMAGES_PREFIX:-}app-php-worker
+  restart: unless-stopped
+  environment:
+    - RUN_MIGRATIONS=false
+  healthcheck:
+    disable: true
+  depends_on:
+    - php
+```
```
7-zete-7, line 14:
```suggestion
```yaml
  php-worker:
    image: ${IMAGES_PREFIX:-}app-php-worker
    restart: unless-stopped
    environment:
      - RUN_MIGRATIONS=false
    healthcheck:
      disable: true
    depends_on:
      - php
```
```

File: docs/symfony-messenger.md [hunk 2]
```diff
@@ -0,0 +1,62 @@
...
+
+Add new service to the `compose.prod.yaml`:
+```yaml
+  php-worker:
+  build:
+    context: .
+    target: frankenphp_prod
+  environment:
+    APP_SECRET: ${APP_SECRET}
+```
```
7-zete-7, line 52:
```suggestion
```yaml
  php-worker:
    build:
      context: .
      target: frankenphp_prod
    environment:
      APP_SECRET: ${APP_SECRET}
```
```

File: docs/symfony-messenger.md [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+
+Apply the following changes to the `frankenphp/docker-entrypoint.sh`:
+```patch
+-	if grep -q ^DATABASE_URL= .env; then
++	run_migrations=${RUN_MIGRATIONS:-true}
++	if grep -q ^DATABASE_URL= .env && [ "$run_migrations" = "true" ]; then
```
7-zete-7, line 58: Maybe combine these lines into one?
```suggestion
+	if grep -q ^DATABASE_URL= .env && [ "${RUN_MIGRATIONS:-true}" = "true" ]; then
```

7-zete-7:
Great work, @k-37!

Do you think it makes sense to create a new entrypoint for the `php-worker` service, so that the `--time-limit`, `--limit` and `--memory-limit` parameters can be set via environment variables, and transports (and additional parameters) via the `command` parameter?

```yaml
   php-worker:
     # ...
     environment:
       - RUN_MIGRATIONS=false
       - CONSUMER_TIME_LIMIT=60
       - CONSUMER_LIMIT=10
       - CONSUMER_MEMORY_LIMIT=128M
     command: ['async']

File: docs/symfony-messenger.md [hunk 4]
```diff
@@ -0,0 +1,62 @@
...
+    - php
+```
+
+Add new services to the `compose.override.yaml`:
+```yaml
+  php-worker:
+    profiles:
+      - donotstart
+
+  php-worker-async:
```
dimonx3, line 22: I don't understand how to run `php-worker-async` service in a prod environment.
Please provide an example of how to add the `php-worker-async` service to the `compose.prod.yaml` file.
Thanks.

dimonx3:
How to run `php-worker-async` service in a **prod** environment?
Please provide an example of how to add the `php-worker-async` service to the `compose.prod.yaml` file.
Thanks.

k-37:
> Do you think it makes sense to create a new entrypoint for the `php-worker` service, so that the `--time-limit`, `--limit` and `--memory-limit` parameters can be set via environment variables, and transports (and additional parameters) via the `command` parameter?

Thank you for your review @7-zete-7. Since we are documenting needed changes to existing project files my intention was to provide minimal needed modifications of existing files, to make it as simple as possible. What do you think?

k-37:
> How to run `php-worker-async` container in a **prod** environment? Please provide an example of how to add the `php-worker-async` service to the `compose.prod.yaml` file.

Thank you @dimonx3 for taking interest in this pull request. I have improved instructions for `compose.prod.yaml`.

docs/symfony-messenger.md
```diff
@@ -44,11 +44,22 @@ To add additional workers just copy `php-worker-async` service and replace every
...
+    extends:
+      file: compose.yaml
+      service: php-worker
+    image: ${IMAGES_PREFIX:-}app-php-worker-async
     build:
       context: .
-      target: frankenphp_prod
-    environment:
-      APP_SECRET: ${APP_SECRET}
+      target: frankenphp_dev
```
dimonx3, line 58: I think target should be frankenphp_prod

k-37: You are right. Thanks.

7-zete-7:
>> Do you think it makes sense to create a new entrypoint for the php-worker service, so that the --time-limit, --limit and --memory-limit parameters can be set via environment variables, and transports (and additional parameters) via the command parameter?
>
> Thank you for your review @7-zete-7. Since we are documenting needed changes to existing project files my intention was to provide minimal needed modifications of existing files, to make it as simple as possible. What do you think?

The idea of ​​adding a separate entrypoint came from how cleverly the default entrypoint was bypassed (using `/app/bin/console` instead of `bin/console`).
https://github.com/dunglas/symfony-docker/blob/dc170d98ebb18518e9266f55c8db2651464ef565/frankenphp/docker-entrypoint.sh#L4

It seems to me that such a trick may not be completely clear. It seems that it is necessary to either describe it or move the launch to a separate entrypoint so that this behavior is more obvious.

7-zete-7:
Perhaps we should move the check for the `RUN_MIGRATIONS` variable to line [51](https://github.com/dunglas/symfony-docker/blob/dc170d98ebb18518e9266f55c8db2651464ef565/frankenphp/docker-entrypoint.sh#L51) instead of [29](https://github.com/dunglas/symfony-docker/blob/dc170d98ebb18518e9266f55c8db2651464ef565/frankenphp/docker-entrypoint.sh#L29)? So that the connection check remains and only migrations are excluded.
https://github.com/dunglas/symfony-docker/blob/dc170d98ebb18518e9266f55c8db2651464ef565/frankenphp/docker-entrypoint.sh#L29-L54

k-37:
@7-zete-7 thank you for your suggestions.

I must admit that I don't quite understand the issue with entrypoint and what improvements should be made.

I'm split on the suggestion to move migration check.

Is it OK to leave both decisions to project maintainers who are more knowledgeable about the project than me?

7-zete-7:
> I must admit that I don't quite understand the issue with entrypoint and what improvements should be made.

The default entrypoint will only work if the first element in command is `frankenphp`, `php`, or `bin/console`.

https://github.com/dunglas/symfony-docker/blob/dc170d98ebb18518e9266f55c8db2651464ef565/frankenphp/docker-entrypoint.sh#L4

This PR uses `/app/bin/console` as the first element in command, which skips entire code of `then` statement of this `if`.

7-zete-7:
> I'm split on the suggestion to move migration check.

In my opinion, when using the `RUN_MIGRATIONS` parameter, it is better to leave the check for a connection to the database, and only disable the application of migrations.

https://github.com/dunglas/symfony-docker/blob/dc170d98ebb18518e9266f55c8db2651464ef565/frankenphp/docker-entrypoint.sh#L51-L53

> Is it OK to leave both decisions to project maintainers who are more knowledgeable about the project than me?

Can you help to choose the appropriate option, @maxhelias?

k-37:
> The default entrypoint will only work if the first element in command is `frankenphp`, `php`, or `bin/console`.

@7-zete-7 thank you for the clarification. What do you think about using:

    command: ['php', '-f', '/app/bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']

instead of:

    command: ['/app/bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']

7-zete-7:
> > The default entrypoint will only work if the first element in command is `frankenphp`, `php`, or `bin/console`.
> 
> @7-zete-7 thank you for the clarification. What do you think about using:
> 
> ```
> command: ['php', '-f', '/app/bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
> ```
> 
> instead of:
> 
> ```
> command: ['/app/bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
> ```

Thanks for the great examples @k-37!

Of these two options, I would prefer the second one.

Although it would make sense to just use `bin/console` instead of `/app/bin/console`.

```
command: ['bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
```

k-37:
`bin/console` it is then. Thank you @7-zete-7.

Regarding your point about `RUN_MIGRATIONS`. Since `php-worker-async` has:

    depends_on:
      php:
        condition: service_healthy

it is started after `php` service which waits for database and runs migrations. `php-worker-async` will start slightly faster without additional database check which is currently used only before running migrations.

My preference would be to leave it as it is. But I'm fine with whatever maintainers like better.

k-37:
Thank you @daFish for taking interest in this PR.

I use `^DATABASE_URL=` [here](https://github.com/k-37/dolphin-is-symfony-ddd-experiment/blob/011eb001af39cf9f47b3dab9d1ff45707029ae50/etc/frankenphp/docker-entrypoint.sh#L29) without quotes around it and it works as expected. I just rebuilt my project to test it.

Here is relevant line in [`.env`](https://github.com/k-37/dolphin-is-symfony-ddd-experiment/blob/011eb001af39cf9f47b3dab9d1ff45707029ae50/.env#L36) file.

Here is [`compose.yaml`](https://github.com/k-37/dolphin-is-symfony-ddd-experiment/blob/011eb001af39cf9f47b3dab9d1ff45707029ae50/compose.yaml#L38) and [`compose.override.yaml`](https://github.com/k-37/dolphin-is-symfony-ddd-experiment/blob/011eb001af39cf9f47b3dab9d1ff45707029ae50/compose.override.yaml#L27).

Do you see any other difference which might be behind your issue?

maxhelias:
Hi @k-37, @7-zete-7,
Thanks for tackle this.
I finally took the time for a first reading. I haven't tested it yes, but it's look a great job.

I'm not really in favor of a new entrypoint to stay as lite as possible. I find that the `time-limit`, `limit` & `memory-limit` options are not necessary in env variable, the command can be easily modified (as we are copy/paste from the doc). WDYT @dunglas ?

For the `docker-entrypoint.sh` modification, the `RUN_MIGRATIONS`, can be apply directly IMO with the default value.

I'll try this soon for a better feedback 😉

docs/symfony-messenger.md
```diff
@@ -0,0 +1,72 @@
...
+    command: ['bin/console', 'messenger:consume', 'async', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M']
+    depends_on:
+      php:
+        condition: service_healthy
+```
+
+Apply the following changes to the `frankenphp/docker-entrypoint.sh`:
+```patch
+-	if grep -q ^DATABASE_URL= .env; then
++	if grep -q ^DATABASE_URL= .env && [ "${RUN_MIGRATIONS:-true}" = "true" ]; then
```
norkunas, line 68: So basically this won't wait anymore for the database to be available?

docs/symfony-messenger.md
```diff
@@ -0,0 +1,72 @@
...
+      disable: true
+    depends_on:
+      - php
+```
+
+Add new services to the `compose.override.yaml`:
+```yaml
+  php-worker:
+    profiles:
+      - donotstart
```
norkunas, line 20: could you explain what's this for? :blush:

docs/symfony-messenger.md
7-zete-7: @norkunas, I agree with you. A similar proposal [was made above](https://github.com/dunglas/symfony-docker/pull/681#issuecomment-2454054046).

[The author refers](https://github.com/dunglas/symfony-docker/pull/681#issuecomment-2468769518) to the fact that the service will only be launched after a healthy `php` service. And this is true.

Although I would abandon the dependency on the `php` service (so that it would be possible to launch the consumer separately from or earlier than the main service), by enabling the wait for the database to be ready. However, in this case, consumers will be launched before the migrations are applied. So for now I agree with the author on the issue of waiting for the database.

docs/symfony-messenger.md
k-37: Hello @norkunas, in `compose.yaml` we have **base** `php-worker` which is extended by `php-worker-async` in `compose.override.yaml` and in `compose.prod.yaml`, `donotstart` is used to prevent separate `php-worker` starting which is only used as a base.

What do you think, is it necessary to add explanation for that in `docs/symfony-messenger.md`?

norkunas: Understood, thanks for explanation :+1: 
I think it could be good to explain for others like me which sees this first time :)

norkunas: Just another question: why not naming these services same? IMHO worker is already basically async because it's another service

k-37: > Just another question: why not naming these services same? IMHO worker is already basically async because it's another service

By using different name it is easy to add additional workers with different configuration. I updated `symony-messenger.md` to try clarify that.

moshkov-konstantin:
@dunglas  Maybe merge?

LaurentSanson:
Hi all,

Any update on this PR? 

Not sure if it's related but my php service is not healthy anymore with the added code

k-37:
Hi @LaurentSanson,
 
Could you please give more details about your service and environment? I would like to reproduce the problem.

If you could provide simplified non working example that would be great.

LaurentSanson:
Hey @k-37 !

Sure, here is some of my actual code : 

```
compose.yaml

services:
    php:
        image: ${IMAGES_PREFIX:-}app-php
        restart: unless-stopped
        environment:
            SERVER_NAME: ${SERVER_NAME:-localhost}, php:80
            MERCURE_PUBLISHER_JWT_KEY: ${CADDY_MERCURE_JWT_SECRET:-!ChangeThisMercureHubJWTSecretKey!}
            MERCURE_SUBSCRIBER_JWT_KEY: ${CADDY_MERCURE_JWT_SECRET:-!ChangeThisMercureHubJWTSecretKey!}
            DATABASE_URL: postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@database:5432/${POSTGRES_DB}?serverVersion=${POSTGRES_VERSION}&charset=${POSTGRES_CHARSET:-utf8}
        volumes:
            - caddy_data:/data
            - caddy_config:/config
        ports:
            # HTTP
            -   target: 80
                published: 80
                protocol: tcp
            # HTTPS
            -   target: 443
                published: 443
                protocol: tcp
            # HTTP/3
            -   target: 443
                published: 443
                protocol: udp
    
    php-worker:
        image: ${IMAGES_PREFIX:-}app-php-worker
        restart: unless-stopped
        environment:
            - RUN_MIGRATIONS=false
        healthcheck:
            disable: true
        depends_on:
            - php
    
    # Mercure is installed as a Caddy module, prevent the Flex recipe from installing another service
    ###> symfony/mercure-bundle ###
    ###< symfony/mercure-bundle ###
    
    ###> doctrine/doctrine-bundle ###
    database:
        image: postgres:${POSTGRES_VERSION:-16}-alpine
        environment:
            POSTGRES_DB: ${POSTGRES_DB:-app}
            # You should definitely change the password in production
            POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-!ChangeMe!}
            POSTGRES_USER: ${POSTGRES_USER:-app}
        healthcheck:
            test: [ "CMD", "pg_isready", "-d", "${POSTGRES_DB:-app}", "-U", "${POSTGRES_USER:-app}" ]
            timeout: 5s
            retries: 5
            start_period: 60s
        volumes:
            - database_data:/var/lib/postgresql/data:rw
            # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!
            # - ./docker/db/data:/var/lib/postgresql/data:rw
    ###< doctrine/doctrine-bundle ###

    ###> sensiolabs/gotenberg-bundle ###
    gotenberg:
        image: 'gotenberg/gotenberg:8'
    ###< sensiolabs/gotenberg-bundle ###

volumes:
    caddy_data:
    caddy_config:
    ###> symfony/mercure-bundle ###
    ###< symfony/mercure-bundle ###
    
    ###> doctrine/doctrine-bundle ###
    database_data:
    ###< doctrine/doctrine-bundle ###
```

```
compose.override.yaml

services:
    php:
        build:
            context: .
            target: frankenphp_dev
        volumes:
            - ./:/app
            - ./frankenphp/Caddyfile:/etc/caddy/Caddyfile:ro
            - ./frankenphp/conf.d/20-app.dev.ini:/usr/local/etc/php/app.conf.d/20-app.dev.ini:ro
            # If you develop on Mac or Windows you can remove the vendor/ directory
            #  from the bind-mount for better performance by enabling the next line:
            #- /app/vendor
        environment:
            MERCURE_EXTRA_DIRECTIVES: demo
            # See https://xdebug.org/docs/all_settings#mode
            XDEBUG_MODE: "${XDEBUG_MODE:-off}"
        extra_hosts:
            # Ensure that host.docker.internal is correctly defined on Linux
            - host.docker.internal:host-gateway
        tty: true
    
    php-worker:
        profiles:
            - donotstart
    php-worker-async:
        scale: 2
        extends:
            file: compose.yaml
            service: php-worker
        image: ${IMAGES_PREFIX:-}app-php-worker-async
        build:
            context: .
            target: frankenphp_dev
        command: [ 'bin/console', 'messenger:consume', 'scheduler_form_analysis', '-vv', '--time-limit=60', '--limit=10', '--memory-limit=128M' ]
        volumes:
            - ./:/app
            - /app/var/
        depends_on:
            php:
                condition: service_healthy
    
    ###> symfony/mercure-bundle ###
    ###< symfony/mercure-bundle ###
    
    ###> doctrine/doctrine-bundle ###
    database:
        ports:
            - "5432:5432"
    ###< doctrine/doctrine-bundle ###
    
    ###> symfony/mailer ###
    maildev:
        image: maildev/maildev
        container_name: maildev-container
        command: bin/maildev --web 1080 --smtp 25 --verbose
        healthcheck:
            test: 'wget -O - http://127.0.0.1:1080/healthz || exit 1'
        ports:
            - "1080:1080"
    ###< symfony/mailer ###

    ###> sensiolabs/gotenberg-bundle ###
    gotenberg:
        ports:
            - "3000"
    ###< sensiolabs/gotenberg-bundle ###
        

```


And here is the error on the CI : 

```
 Network testmyform_default  Creating
 Network testmyform_default  Created
 Volume "testmyform_caddy_data"  Creating
 Volume "testmyform_caddy_data"  Created
 Volume "testmyform_caddy_config"  Creating
 Volume "testmyform_caddy_config"  Created
 Volume "testmyform_database_data"  Creating
 Volume "testmyform_database_data"  Created
 Container maildev-container  Creating
 Container testmyform-database-1  Creating
 Container testmyform-php-1  Creating
 Container testmyform-gotenberg-1  Creating
 Container testmyform-gotenberg-1  Created
 Container testmyform-database-1  Created
 Container maildev-container  Created
 Container testmyform-php-1  Created
 Container testmyform-php-worker-async-2  Creating
 Container testmyform-php-worker-async-1  Creating
 Container testmyform-php-worker-async-1  Created
 Container testmyform-php-worker-async-2  Created
 Container maildev-container  Starting
 Container testmyform-php-1  Starting
 Container testmyform-database-1  Starting
 Container testmyform-gotenberg-1  Starting
 Container testmyform-gotenberg-1  Started
 Container testmyform-database-1  Started
 Container maildev-container  Started
 Container testmyform-php-1  Started
 Container testmyform-php-1  Waiting
 Container testmyform-php-1  Error
dependency failed to start: container testmyform-php-1 is unhealthy
Error: Process completed with exit code 1.
```

I hope this will help, but once again not 100% sure it's related to the added code but kind of

k-37:
@LaurentSanson,

Can you SSH into your CI server and get output of the command:

    docker logs testmyform-php-1
    
Do you have any problems when running your application with Docker outside of CI, i.e. locally?

LaurentSanson:
Locally I forgot to add the `--wait` option... I've just added it and rebuilt all the containers, here is the error I have : 

`container xxx-php-worker-async-1 has no healthcheck configured`

k-37:
That looks like a separate issue maybe caused by `--wait` option, which waits for services to be running|healthy.

In `compose.yaml` you have:

    php-worker:
        ...
        healthcheck:
            disable: true

Health check is disabled for `php-worker` because worker will restart itself periodically.

Your original error is about `php` container not about `php-worker`.

Try starting your application locally with these commands and report eventual errors:

    docker compose down --remove-orphans
    docker compose build --pull --no-cache
    docker compose up # Note that we are not using `--detach` option, we want to see logs in the terminal

I have [working example of application](https://github.com/k-37/dolphin-is-symfony-ddd-experiment) using async Symfony messenger. Your compose files look pretty similar to mine. `Container testmyform-php-1  Error` might be something specific to your application or environment.

LaurentSanson:
Yes if the containers are restarting periodically, we can't use the `--wait` option anymore I guess... that's quite a shame but that'll do for now

> I have [working example of application](https://github.com/k-37/dolphin-is-symfony-ddd-experiment) using async Symfony messenger. Your compose files look pretty similar to mine. `Container testmyform-php-1 Error` might be something specific to your application or environment.

Yes I found an error due to migrations, only related to my mistakes, my bad.

docs/symfony-messenger.md
```diff
@@ -0,0 +1,74 @@
...
+
+  php-worker-async:
+    scale: 2
```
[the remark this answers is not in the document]
7-zete-7: Can you share information where it was announced that this attribute became deprecated?

This attribute is not deprecated in the specification (yet?)

See:
* [compose-spec/compose-spec/05-services.md](https://github.com/compose-spec/compose-spec/blob/00655e2222608fa6eaecdf12ac5195889ca2244a/05-services.md#scale)
* [compose-spec/compose-spec/schema/compose-spec.json](https://github.com/compose-spec/compose-spec/blob/00655e2222608fa6eaecdf12ac5195889ca2244a/schema/compose-spec.json#L717-L720)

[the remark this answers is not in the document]
norkunas: firstly chat gpt said this to me, then i checked
https://github.com/docker/compose/issues/10823
https://devops.stackexchange.com/questions/6233/how-to-create-replicas-in-docker-compose-when-scale-is-deprecated

if i'm wrong, then i am happy :grinning:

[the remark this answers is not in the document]
7-zete-7: > https://github.com/docker/compose/issues/10823

This is about the `docker compose scale ...` command.

> https://devops.stackexchange.com/questions/6233/how-to-create-replicas-in-docker-compose-when-scale-is-deprecated

This is about incompatibility of Docker Compose configuration migration from V2 to V3. This has been fixed (see https://github.com/docker/compose/issues/5586#issuecomment-847221446).

HectorOrdonez:
Such a pity this was left unfinished. 

@k-37 is this something you plan to pick up, or is it an idea that someone else continues it?

k-37:
Hello @HectorOrdonez, I consider it finished, I have addressed all of the comments from PR.

Its destiny is up to maintainers.

PR #686, opened 2024-11-10
Add LICENSE

anzweidrej:


dunglas:
I don't think we should merge this because the repository is a template, and the file will be included in rejects created from this template, which are likely not MIT-licensed.

dunglas:
According to https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#determining-the-location-of-your-license, a mention in the README is good enough, even if having a LICENSE file is a best practice.

maxhelias:
We have this mention here : https://github.com/dunglas/symfony-docker#license

PR #689, opened 2024-11-11
Fix missing envvar escape

dario-nuro:
Allows new install after cloning.


maxhelias:
Using `${SERVER_NAME}` substitutes the value immediately in Docker Compose, while `$${SERVER_NAME}` delays substitution until runtime inside the container. In your case, the final value by default will be : `https://localhost, php:80/.well-known/mercure` and it's not a good value.

dario-nuro:
Thanks for checking and sorry for rushing into an improper solution, but it allowed me to install.

PR #696, opened 2024-11-16
fix: do not handle requests to Mercure as PHP routes

### LINKED ISSUE #695: Requests to Mercure are handled using PHP
Requests to Mercure are handled with PHP instead of Mercure

This is because after merge #694, all requests except static are redirected to PHP

https://github.com/dunglas/symfony-docker/blob/12944cb6d5f470a45faca7ff9411a0abb190f0e3/frankenphp/Caddyfile#L47

bkosun:
Fix #695

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
         run: curl -vk --fail-with-body https://localhost
       -
         name: Check Mercure reachability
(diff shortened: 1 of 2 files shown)


MrAns0 (approved):

senzowayne:
thanks for the fix 😄

dunglas:
Thank you very much @bkosun (I made some slight tweaks for performance).

PR #712, opened 2024-12-26
prevent direct access to index.php

yassinefikri:


bkosun:
@yassinefikri, we could use the `hide` parameter of the `file_server` directive to hide system files.

https://caddyserver.com/docs/caddyfile/directives/file_server#hide

See #714 

It seems to be a simpler solution. In addition, it has less impact on performance.

WDYT?

bkosun (commented):
@yassinefikri, we could use the `hide` parameter of the `file_server` directive to hide system files.

https://caddyserver.com/docs/caddyfile/directives/file_server#hide

See #714 

It seems to be a simpler solution. In addition, it has less impact on performance.

WDYT?

PR #714, opened 2025-01-03
fix: prevent access to source code of app index file

### LINKED ISSUE #713: Source code of app index file can be accessed from outside
The current configuration of Caddy allows access to the source code of the `index.php` file, this can give an attacker information about the technology stack (framework) being used.

https://github.com/dunglas/symfony-docker/blob/5a973670e1ab869f0531fc10b9cbe82229736d9b/frankenphp/Caddyfile#L56

Such a leak can be used for more targeted attacks, such as finding undocumented vulnerabilities or bypassing existing defenses.

Related PR: #712, #714

bkosun:
Fix #713

### DIFF
--- a/frankenphp/Caddyfile
+++ b/frankenphp/Caddyfile
@@ -53,5 +53,7 @@
 	@frontController path index.php
 	php @frontController
 
-	file_server
+	file_server {
+		hide index.php
+	}
 }


frankenphp/Caddyfile
```diff
@@ -53,5 +53,7 @@
 	@frontController path index.php
 	php @frontController
 
-	file_server
+	file_server {
+		hide index.php
```
dunglas, line 57:
```suggestion
		hide *.php
```

Just in case?

bkosun: Wouldn't this have a negative impact on performance if there are a lot of static files?

bkosun: And it seems we should use `hide /index.php` or `hide /*.php` instead of  `hide index.php` or `hide *.php`

> Specifying a file name or pattern **without a path separator** will hide all files with a matching name regardless of its location; otherwise, a path prefix match will be attempted, and then a globular match
https://caddyserver.com/docs/caddyfile/directives/file_server#hide

WDYT?

bkosun: And it seems we should use `hide public/index.php` instead of `hide index.php`

> Specifying a file name or pattern **without a path separator will hide all files with a matching name** regardless of its location; otherwise, a path prefix match will be attempted, and then a globular match

https://caddyserver.com/docs/caddyfile/directives/file_server#hide

I had to specify the full file path in the `hide public/index.php` code because the` hide /index.php` code didn't work, even though the root directory of the file server must match the main root directory ({http.vars.root})

https://github.com/dunglas/symfony-docker/blob/5a973670e1ab869f0531fc10b9cbe82229736d9b/frankenphp/Caddyfile#L22

> root sets the path to the site root. It's similar to the root directive except it applies to this file server instance only and overrides any other site root that may have been defined. **Default: {http.vars.root} or the current working directory**. Note: This subdirective only changes the root for this handler. For other directives (like try_files or templates) to know the same site root, use the root directive instead.

https://caddyserver.com/docs/caddyfile/directives/file_server#root

WDYT?

maxhelias:
Thanks @bkosun !

PR #715, opened 2025-01-03
Update 20-app.prod.ini

94noni:
hello,

as the page symfony performance [is linked here](https://github.com/dunglas/symfony-docker/blob/5a973670e1ab869f0531fc10b9cbe82229736d9b/frankenphp/conf.d/10-app.ini#L7)
when reading it I saw also this `opcache.validate_timestamps = 0` config

do you think it is worth adding it here as well?

thank you,


maxhelias:
Thanks @94noni !

PR #717, opened 2025-01-09
feat: activate the watch mode in dev

94noni:
hey

the blog post https://dunglas.dev/2024/11/frankenphp-1-3-massive-performance-improvements-watcher-mode-dedicated-prometheus-metrics-and-more/ 
the paragraph "File Watchers" talks about this feature `watch`

Questions:

1. is it good here? no overhead for `prod`
2. otherwise, how to differentiate a `worker.Caddyfile` from `dev` or `prod` ?
    - creating a `ENV FRANKENPHP_CONFIG="import dev.worker.Caddyfile"` inside the `frankenphp_dev`
3. actually what is the best way with current image/container to see live file changes of symfony app?
    - with running `docker compose up --watch` after `docker compose build --no-cache --pull` ?

thank you,

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,6 +54,7 @@ CMD [ "frankenphp", "run", "--config", "/etc/caddy/Caddyfile" ]
 FROM frankenphp_base AS frankenphp_dev
 
 ENV APP_ENV=dev XDEBUG_MODE=off
+ENV FRANKENPHP_CONFIG="import dev.worker.Caddyfile"
 
 RUN mv "$PHP_INI_DIR/php.ini-development" "$PHP_INI_DIR/php.ini"
 
@@ -63,6 +64,7 @@ RUN set -eux; \
 	;
 
 COPY --link frankenphp/conf.d/20-app.dev.ini $PHP_INI_DIR/app.conf.d/
+COPY --link frankenphp/dev.worker.Caddyfile /etc/caddy/dev.worker.Caddyfile
 
 CMD [ "frankenphp", "run", "--config", "/etc/caddy/Caddyfile", "--watch" ]
 
--- a/frankenphp/dev.worker.Caddyfile
+++ b/frankenphp/dev.worker.Caddyfile
@@ -1,0 +1,5 @@
+worker {
+	file ./public/index.php
+	env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime
+	watch
+}


maxhelias (approved):

dunglas:
The watch mode should only be enabled en dev. It should be useless in prod.

94noni:
yes that was my first idea, will see to change it thx

94noni:
@dunglas @maxhelias i am really unsure of the way to achieve this
does this works? or perhaps something like so ?

```yaml
worker {
	file ./public/index.php
	env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime
	{$WATCH_MODE}
}
```

Dockerfile
```diff
@@ -63,6 +64,7 @@ RUN set -eux; \
 	;
 
 COPY --link frankenphp/conf.d/20-app.dev.ini $PHP_INI_DIR/app.conf.d/
+COPY --link frankenphp/dev.worker.Caddyfile /etc/caddy/dev.worker.Caddyfile
 
 CMD [ "frankenphp", "run", "--config", "/etc/caddy/Caddyfile", "--watch" ]
```
deluxetom, line 69: frankenphp is already running in `watch` mode for dev, if you add the worker for dev (which I think should be in your own `Dockerfile`, not the one in this repo) it shouldn't run in watch mode here

deluxetom: I dont think its useless but I think what you're trying to do it out of scope of this repository.
Worker mode isn't meant to be used for development, and watch mode isn't supposed to be used on prod.

What you might want to do is to fork this repository and change the Dockerfile for your own application. This is what I ended up doing on my end.

Trim: @deluxetom is right: the Dockerfile in the symfony-docker project already manages the watch and worker modes.

The Dockerfile builds the project with 3 targets:

1. The first one, the [frankenphp_base](https://github.com/dunglas/symfony-docker/blob/6707f3d1c06d45062eca1afd8693c6e9e57e5a0f/Dockerfile#L12) build target, is a common target for development and production image. It contains all dependencies which are common to both type of environement
2. The second one is [frankenphp_dev](https://github.com/dunglas/symfony-docker/blob/6707f3d1c06d45062eca1afd8693c6e9e57e5a0f/Dockerfile#L54) which takes the base one, adds the debugger and [add the `--watch` argument to the `frankenphp` command](https://github.com/dunglas/symfony-docker/blob/6707f3d1c06d45062eca1afd8693c6e9e57e5a0f/Dockerfile#L67)
3. The third one is [frankenphp_prod](https://github.com/dunglas/symfony-docker/blob/6707f3d1c06d45062eca1afd8693c6e9e57e5a0f/Dockerfile#L70) which is independent from the dev one (so the `--watch` argument is not used) and which downloads dependencies and enables the `worker` mode through the line [`ENV FRANKENPHP_CONFIG="import worker.Caddyfile"`](https://github.com/dunglas/symfony-docker/blob/6707f3d1c06d45062eca1afd8693c6e9e57e5a0f/Dockerfile#L73C5-L73C22) (this environment variable is read by the main Caddyfile; the `worker.Caddyfile` enables the worker mode)

It is well designed by default, but it can be forked and adapted as you prefer as @deluxetom said

94noni: oki thx both of you with all your comments
of course after using this repo as base I've tweaked some file and perhaps missed something regarding such watch logic as when testing as I had to rebuild to have y code available
will double check this specifically 

in the meantime lets close here if it wont bring value as it should be already the case

thx again !

PR #728, opened 2025-01-28
fix (Makefile): Configure editor to use tabs by default

mfe-a5sys:
Add indent style to tabs when editing Makefiles.

Makefiles does not work with spaces.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -55,3 +55,6 @@ indent_style = tab
 
 [*.*Caddyfile]
 indent_style = tab
+
+[Makefile]
+indent_style = tab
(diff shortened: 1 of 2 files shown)


7-zete-7:
Hi @mfe-a5sys!

This is a useful fix! However, I think it should be described for the documentation file [`docs/makefile.md`](https://github.com/dunglas/symfony-docker/blob/main/docs/makefile.md), and not in the `.editorconfig` file. This template does not provide work with make by default. Only after making changes according to the documentation file.

mfe-a5sys:
Ah right ! I will edit this file then, thanks

maxhelias (approved):

maxhelias:
Thanks @mfe-a5sys !

PR #733, opened 2025-02-09
Fix typo in xdebug.md

delyro:


docs/xdebug.md
```diff
@@ -15,7 +15,7 @@ XDEBUG_MODE=debug docker compose up -d
 On Windows:
 
 ```
-set XDEBUG_MODE=debug&& docker compose up -d&set XDEBUG_MODE=
+set XDEBUG_MODE=debug && docker compose up -d
```
maxhelias, line 18: I think we need to keep the reset after the command to reduce side effects, as windows keeps the environment variable during the cmd session.

delyro: You're right! Another issue is that the default terminal on Windows is PowerShell, so the documentation won't work as written, and we need to use cmd for this to function correctly. I will close the pull request.

PR #738, opened 2025-02-13
Update tls.md

abdounikarim:
When using a custom TLS certificate, we also need to update the `SERVER_NAME` env variable.

### DIFF
--- a/docs/tls.md
+++ b/docs/tls.md
@@ -31,6 +31,7 @@ For instance, to use self-signed certificates created with [mkcert](https://gith
     php:
       environment:
     +    CADDY_SERVER_EXTRA_DIRECTIVES: "tls /etc/caddy/certs/tls.pem /etc/caddy/certs/tls.key"
+    +    SERVER_NAME: server-name.localhost"
         # ...
       volumes:
     +    - ./frankenphp/certs:/etc/caddy/certs:ro


7-zete-7:
Hi @abdounikarim!

Thanks for suggesting this change!

This environment variable is already set in _compose.yaml_. No need to set it in _compose.override.yaml_.

It makes more sense to describe that the Docker Compose project should be launched with this environment variable explicitly set:

- Launch the Docker Compose project with a command like this:
    ```bash
    SERVER_NAME=server-name.localhost \
    docker compose ...
    ```

- Or set the value of this environment variable in the _.env_ file:
    ```diff
    +SERVER_NAME=server-name.localhost
    ```

- Or replace the default value for all forwardings of the `SERVER_NAME` environment variable in _compose.yaml_ files with the required one:
    ```diff
    -SERVER_NAME: ${SERVER_NAME:-localhost}, php:80
    +SERVER_NAME: ${SERVER_NAME:-server-name.localhost}, php:80
    ```

abdounikarim:
Hi @7-zete-7 and thanks for your answer.

Your partially right actually 😉
We can use the env variable defined in `compose.yaml`, but we need to:
- remove the `php:80` at the end or
- change it to `php:443`

If you don't, you'll have this error:

```
Error: adapting config using caddyfile: server listening on [:80] is HTTP, but attempts to configure TLS connection policies
```

7-zete-7:
Thanks for the feedback, @abdounikarim!

Yes, indeed, if you leave `php:80`, you will get the error you described. Forgot about that 😅

Perhaps the best option would be to completely abandon the `php` hostname. The custom TLS certificate, I assume, will not support the `php` hostname. Because of this, calls to FrankenPHP via the `php` hostname will result in a TLS error.

In the default configuration, the `php` hostname is used only for Symfony to call Mercure locally. You can simply configure access to Mercure via an external hostname:
```diff
 services:
   php:
     # ...
     environment:
       # ...
-      SERVER_NAME: ${SERVER_NAME:-localhost}, php:80
+      SERVER_NAME: ${SERVER_NAME:-server-name.localhost}
       # ...
-      MERCURE_URL: ${CADDY_MERCURE_URL:-http://php/.well-known/mercure}
+      MERCURE_URL: ${CADDY_MERCURE_URL:-https://${SERVER_NAME:-server-name.localhost}/.well-known/mercure}
```
In this case, there will be no need for the hostname `php` at all.

In addition, for optimization (excluding access via more network nodes than necessary), you can also set an additional hostname for service `php` (see https://docs.docker.com/reference/compose-file/services/#aliases):
```diff
 services:
   php:
     # ...
+    networks:
+      app:
+        aliases:
+          - ${SERVER_NAME:-server-name.localhost}
 # ...
+networks:
+  app:
```
After such optimization, requests will pass through the same number of network nodes as in the standard configuration.

maxhelias:
Discussion about it here : https://github.com/dunglas/symfony-docker/pull/774

Issue #751, opened 2025-03-07
Display of phpinfo() output in production


LinkNexus:
Hello! I'm trying to deploy my web-app by following the documentation from this repo. In local, everything works fine, without problem but when i try to deploy the app by following the documentation of this repo, some default page is displayed instead of my website, the default page being the output of phpinfo() function. 
The repo of the project is here: https://github.com/[LinkNexus/GamersWorld](https://github.com/LinkNexus/GamersWorld).
Here is also the content of my Dockerfiles:

```
#Dockerfile
#syntax=docker/dockerfile:1

# Versions
FROM dunglas/frankenphp:1-php8.3 AS frankenphp_upstream

# The different stages of this Dockerfile are meant to be built into separate images
# https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage
# https://docs.docker.com/compose/compose-file/#target

# Base FrankenPHP image
FROM frankenphp_upstream AS frankenphp_base

WORKDIR /app

VOLUME /app/var/

# persistent / runtime deps
# hadolint ignore=DL3008
RUN apt-get update && apt-get install -y --no-install-recommends \
    acl \
    file \
    gettext \
    git \
    curl \
    && rm -rf /var/lib/apt/lists/*

RUN set -eux; \
    install-php-extensions \
        @composer \
        apcu \
        intl \
        opcache \
        zip \
        gd \
        zlib \
    ;

# https://getcomposer.org/doc/03-cli.md#composer-allow-superuser
ENV COMPOSER_ALLOW_SUPERUSER=1

ENV PHP_INI_SCAN_DIR=":$PHP_INI_DIR/app.conf.d"

###> recipes ###
###> doctrine/doctrine-bundle ###
RUN install-php-extensions pdo_pgsql
###< doctrine/doctrine-bundle ###
###< recipes ###

COPY --link frankenphp/conf.d/10-app.ini $PHP_INI_DIR/app.conf.d/
COPY --link --chmod=755 frankenphp/docker-entrypoint.sh /usr/local/bin/docker-entrypoint
COPY --link frankenphp/Caddyfile /etc/caddy/Caddyfile

ENTRYPOINT ["docker-entrypoint"]

HEALTHCHECK --start-period=60s CMD curl -f http://localhost:2019/metrics || exit 1
CMD [ "frankenphp", "run", "--config", "/etc/caddy/Caddyfile" ]

# Dev FrankenPHP image
[...]

Issue #745, opened 2025-02-20
#prod home route is redirected to some php Info default web page


upgrader-dev:
Hello, 

I have an existing project and followed the documentation regarding this case.

Everything works just fine in dev mode on my local machine.

When I deploy on my server in production mode I get no error but when I get to my server name, I am being redirected to some default php info Web page and never on my app.

Same error than here apparently : 
https://stackoverflow.com/questions/77761841/symfony-ignores-routing-and-shows-phpinfo

I also tried to deploy the app on my server in dev mode but same issue.

Any idea?

7-zete-7:
Hi @upgrader-dev!

It's hard to say what could be the reason based on the current information.

It looks like this is related to https://github.com/dunglas/symfony-docker/issues/632.

Will you be able to create a minimal reproducer of this issue? Or a sequence of actions that can lead to this error?

upgrader-dev:
Exact same issue you're right. But despite its status "Closed", I was unable to find a solution in this one https://github.com/dunglas/symfony-docker/issues/632

To create this issue I followed the same steps as the issue above.

7-zete-7:
@upgrader-dev, can you try to build images and then restart the Docker Compose project and see if the problem persists?

I suspect that you are running images that do not contain your project files (this applies to the prod environment). Rebuilding the images should solve this problem.

Build new images for dev and prod environments (with explicitly disabling the build cache):
```shell
docker compose -f compose.yaml -f compose.override.yaml build --no-cache php
docker compose -f compose.yaml -f compose.prod.yaml build --no-cache php
```

Can you also clarify how exactly you tried to deploy the app in dev mode?

upgrader-dev:
Thank you for your help @7-zete-7 .
I have tried many times rebuilding fresh images with --no-cache option without success.

I confirm that my server contains my project files as I can see them with visual studio code of a terminal running ssh.

I builded new images as you wrote above but when I deploy I get "container php is unhealthy".

To build the app in dev mode, I run
```
SERVER_NAME=your-domain-name.example.com \
APP_SECRET=ChangeMe \
CADDY_MERCURE_JWT_SECRET=ChangeThisMercureHubJWTSecretKey \
docker compose -f compose.yaml up -d --wait
```
My thoughts were : As the docker compose.yaml works on my local machine, there must be an issue with compose.prod.yaml

7-zete-7:
Thanks for the info @upgrader-dev!

[...]

PR #752, opened 2025-03-07
doc: add information about building production image

tcoch:
Due to issues #745 (probably) and #751 (and some seens on StackOverflow also), I propose to be a little more precise about the building of production images.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -16,7 +16,10 @@ Remember that, by default, if you run `docker compose up -d`, only the files `co
 See https://docs.docker.com/compose/intro/compose-application-model.
 
 If you need to build images for production environment, you have to use the following command:
-`docker compose -f compose.yaml -f compose.prod.yaml build --no-cache`.
+
+```console
+docker compose -f compose.yaml -f compose.prod.yaml build --no-cache
+```
 
 ### Why application outputs `phpinfo()`
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -58,7 +58,11 @@ This means you have to run: `docker compose -f compose.yaml -f compose.prod.yaml
 
 That way, you will see that `frankenphp_base` and `frankenphp_prod` are built this time, which is what you will need for production purposes.
 
-You can finally start your prod container(s) by running: `docker compose -f compose.yaml -f compose.prod.yaml up -d`.
+You can finally start your prod container(s) by running:
+
+```console
+docker compose -f compose.yaml -f compose.prod.yaml up -d
+```
 
 </details>
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -7,3 +7,128 @@ If you work on linux and cannot edit some of the project files right after the f
 ## TLS/HTTPS Issues
 
 See more in the [TLS section](tls.md)
+
+## Production issues
+
+### How to properly build fresh images for production use
+
+Remember that, by default, if you run `docker compose up -d`, only the files `compose.yaml` and `compose.override.yaml` will be used.
+See https://docs.docker.com/compose/intro/compose-application-model.
+
+If you need to build images for production environment, you have to use the following command:
+`docker compose -f compose.yaml -f compose.prod.yaml build --no-cache`.
+
+### Why application outputs `phpinfo()`
+
+Both dev and prod images have the same image tag (`<...>app-php:latest`). This can cause confusion when working with images.
+It is important to make sure that your image is the appropriate one for the current environment.
+
+If you are not careful about this, and try to run your production container(s) with
+`docker compose -f compose.yaml -f compose.prod.yaml up -d`
+without the right build process beforehand, your application **will still launch**, but will be displaying an output of `phpinfo()` (or possibly even a HTTP 500 error page).
+
+See details below.
+
+<details>
+
+<summary>Output of a basic build process</summary>
+
+In the case of a dev image, you need the `compose.yaml` and `compose.override.yaml` files. Which are the default files for Docker Compose.
+
+This means that running `docker compose <command>` or `docker compose -f compose.yaml -f compose.override.yaml <command>` is the same thing.
+
+```
+> docker compose build --no-cache
+[+] Building 287.5s (21/21) FINISHED                                                                                                                                                        docker:desktop-linux
+=> [php internal] load build definition from Dockerfile                                                                                                                                                    0.0s
+=> => transferring dockerfile: 2.96kB                                                                                                                                                                      0.0s
+=> [php] resolve image config for docker-image://docker.io/docker/dockerfile:1                                                                                                                             1.7s
+=> [php auth] docker/dockerfile:pull token for registry-1.docker.io                                                                                                                                        0.0s
+=> CACHED [php] docker-image://docker.io/docker/dockerfile:1@sha256:4c68376a702446fc3c79af22de146a148bc3367e73c25a5803d453b6b3f722fb                                                                       0.0s
+=> [php internal] load metadata for docker.io/dunglas/frankenphp:1-php8.3                                                                                                                                  0.7s
+=> [php auth] dunglas/frankenphp:pull token for registry-1.docker.io                                                                                                                                       0.0s
+=> [php internal] load .dockerignore                                                                                                                                                                       0.0s
+=> => transferring context: 498B                                                                                                                                                                           0.0s
+=> [php frankenphp_upstream 1/1] FROM docker.io/dunglas/frankenphp:1-php8.3@sha256:80a0db5e0b3bec1c93067407fe0a9047e6af41202d5e476c0a3c2101238ce4f6                                                        0.0s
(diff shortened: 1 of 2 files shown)


7-zete-7:
Hi, @tcoch!

Thank you for adding this information to the manual. This is really important information!

It seems to me that it is better not to place the instructions on how the project should be built for deployment in the _Deployment_ section.

Perhaps it would be worth adding an additional section like _Preparing for deployment_?

Then, in addition to clarifying the command for building the image, it would be possible to list some nuances when working with images of this template.

tcoch:
Hi @7-zete-7,
I kinda like the `production.md` file as it is, I don't want it to have too much informations.
I'm thinking maybe I could add these extra informations to the `troubleshooting.md` file ? WDYT ?

7-zete-7:
Sounds very reasonable, @tcoch!

You are right, it will be better than my suggestion. And the `production.md` file will not be overloaded with information.

docs/troubleshooting.md
```diff
@@ -7,3 +7,128 @@ If you work on linux and cannot edit some of the project files right after the f
...
+
+</details>
+
+<br>
+
+<details>
+
+<summary>Output expected for the production build process</summary>
+
+Start by building the image production.
```
uglykidmat, line 87: "Production image" ?

docs/troubleshooting.md
```diff
@@ -7,3 +7,128 @@ If you work on linux and cannot edit some of the project files right after the f
...
+ => [php frankenphp_base 4/7] RUN install-php-extensions pdo_pgsql                                                                                                                                         26.3s
+ => [php frankenphp_base 5/7] COPY --link frankenphp/conf.d/10-app.ini /usr/local/etc/php/app.conf.d/                                                                                                       0.0s
+ => [php frankenphp_base 6/7] COPY --link --chmod=755 frankenphp/docker-entrypoint.sh /usr/local/bin/docker-entrypoint                                                                                      0.0s
+ => [php frankenphp_base 7/7] COPY --link frankenphp/Caddyfile /etc/caddy/Caddyfile                                                                                                                         0.0s
+ => [php frankenphp_prod 1/9] RUN mv "/usr/local/etc/php/php.ini-production" "/usr/local/etc/php/php.ini"                                                                                                   0.2s
```
maxhelias, line 110: Can we get rid of all the logs? And keep only a line to explain where to find the information?

maxhelias (commented):
Thanks for this contribution, I just have one question

docs/troubleshooting.md
tcoch: You got it ;)

docs/troubleshooting.md
```diff
@@ -7,3 +7,55 @@ If you work on linux and cannot edit some of the project files right after the f
 ## TLS/HTTPS Issues
 
 See more in the [TLS section](tls.md)
+
+## Production issues
+
+### How to properly build fresh images for production use
+
+Remember that, by default, if you run `docker compose up -d`, only the files `compose.yaml` and `compose.override.yaml` will be used.
+See https://docs.docker.com/compose/intro/compose-application-model.
```
7-zete-7, line 16: What do you think, @tcoch, would it make sense to also add a link to information about how Docker Compose config file merge works?
```suggestion
See https://docs.docker.com/compose/intro/compose-application-model and https://docs.docker.com/compose/how-tos/multiple-compose-files/merge.
```

docs/troubleshooting.md
```diff
@@ -7,3 +7,55 @@ If you work on linux and cannot edit some of the project files right after the f
...
+
+If you need to build images for production environment, you have to use the following command:
+`docker compose -f compose.yaml -f compose.prod.yaml build --no-cache`.
```
7-zete-7, line 19: To maintain a common command format in documents

````suggestion
If you need to build images for production environment, you have to use the following command:

```console
docker compose -f compose.yaml -f compose.prod.yaml build --no-cache
```
````

<br>

Example
https://github.com/dunglas/symfony-docker/blob/65676b92b30cf6dc4bf89dd967ea14ee659fc93f/docs/production.md?plain=1#L82-L89

File: docs/troubleshooting.md [hunk 5]
```diff
@@ -7,3 +7,55 @@ If you work on linux and cannot edit some of the project files right after the f
...
+
+### Why application outputs `phpinfo()`
+
+Both dev and prod images have the same image tag (`<...>app-php:latest`). This can cause confusion when working with images.
+It is important to make sure that your image is the appropriate one for the current environment.
+
+If you are not careful about this, and try to run your production container(s) with
+`docker compose -f compose.yaml -f compose.prod.yaml up -d`
+without the right build process beforehand, your application **will still launch**, but will be displaying an output of `phpinfo()` (or possibly even a HTTP 500 error page).
```
7-zete-7, line 28:
````suggestion
If you are not careful about this, and try to run your production container(s) with

```console
docker compose -f compose.yaml -f compose.prod.yaml up -d
```

without the right build process beforehand, your application **will still launch**, but will be displaying an output of `phpinfo()` (or possibly even a HTTP 500 error page).
````

File: docs/troubleshooting.md [hunk 6]
```diff
@@ -7,3 +7,55 @@ If you work on linux and cannot edit some of the project files right after the f
...
+
+See details below.
+
+<details>
+
+<summary>Output of a basic build process</summary>
+
+In the case of a dev image, you need the `compose.yaml` and `compose.override.yaml` files. Which are the default files for Docker Compose.
+This means that running `docker compose <command>` or `docker compose -f compose.yaml -f compose.override.yaml <command>` is the same thing.
```
7-zete-7, line 37:
````suggestion
In the case of a dev image, you need the `compose.yaml` and `compose.override.yaml` files. Which are the default files for Docker Compose.
This means that running

```console
docker compose <command>
```

or

```console
docker compose -f compose.yaml -f compose.override.yaml <command>
```

is the same thing.
````

File: docs/troubleshooting.md [hunk 7]
```diff
@@ -7,3 +7,55 @@ If you work on linux and cannot edit some of the project files right after the f
...
+</details>
+
+<br>
+
+<details>
+
+<summary>Output expected for the production build process</summary>
+
+To build the production image, you <ins>have to</ins> specify the `compose.yaml` and `compose.prod.yaml` files.
+This means you have to run: `docker compose -f compose.yaml -f compose.prod.yaml up -d` in order to build your image
```
7-zete-7, line 56:
````suggestion
This means you have to run:

```console
docker compose -f compose.yaml -f compose.prod.yaml up -d
```

in order to build your image
````

File: docs/troubleshooting.md [hunk 8]
```diff
@@ -7,3 +7,55 @@ If you work on linux and cannot edit some of the project files right after the f
...
+(careful: the order of `-f` arguments is important).
+
+That way, you will see that `frankenphp_base` and `frankenphp_prod` are built this time, which is what you will need for production purposes.
+
+You can finally start your prod container(s) by running: `docker compose -f compose.yaml -f compose.prod.yaml up -d`.
```
7-zete-7, line 61:
````suggestion
You can finally start your prod container(s) by running:

```console
docker compose -f compose.yaml -f compose.prod.yaml up -d
```
````

File: docs/troubleshooting.md (hunk 6 above)
Comment (tcoch) [reply docs/troubleshooting.md:40]: Since this part is just an technical explanation and not the end result, I'd like to keep the actual format, which does not highlight the commands too much (so as to not provide too much informations at first glance)

File: docs/troubleshooting.md [hunk 9]
```diff
@@ -7,3 +7,62 @@ If you work on linux and cannot edit some of the project files right after the f
...
+
+<summary>Output of a basic build process</summary>
+
+In the case of a dev image, you need the `compose.yaml` and `compose.override.yaml` files. Which are the default files for Docker Compose.
+This means that running `docker compose <command>` or `docker compose -f compose.yaml -f compose.override.yaml <command>` is the same thing.
+
+And in doing so, images `frankenphp_base` and `frankenphp_dev` are built. And not `frankenphp_prod`.
+Which is good enough for dev purposes.
+
+Then, you can start your dev container(s) by running: `docker compose up -d`.
```
maxhelias, line 45:
````suggestion
Then, you can start your dev container(s) by running:

```console
docker compose up -d
```

````
To match the block above

File: docs/troubleshooting.md [hunk 10]
```diff
@@ -7,3 +7,67 @@ If you work on linux and cannot edit some of the project files right after the f
...
+</details>
+
+<br>
+
+<details>
+
+<summary>Output expected for the production build process</summary>
+
+To build the production image, you <ins>have to</ins> specify the `compose.yaml` and `compose.prod.yaml` files.
+This means you have to run: `docker compose -f compose.yaml -f compose.prod.yaml up -d` in order to build your image
```
7-zete-7, line 61: Fix to _build_ command, not _run_ command
```suggestion
This means you have to run: `docker compose -f compose.yaml -f compose.prod.yaml build` in order to build your image
```

7-zete-7: Does it make sense to add the `--no-cache` flag here?

maxhelias (approved):

PR #765, opened 2025-03-26
feat: enable workers in dev (watchers)

dunglas:
Use the new "watcher" feature of FrankenPHP 1.4 to enable the worker mode in dev. Workers are automatically restarted when a file change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,9 @@ CMD [ "frankenphp", "run", "--config", "/etc/caddy/Caddyfile" ]
 # Dev FrankenPHP image
 FROM frankenphp_base AS frankenphp_dev
 
-ENV APP_ENV=dev XDEBUG_MODE=off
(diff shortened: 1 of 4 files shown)


deluxetom (approved):
much cleaner this way, thanks!

Trim:
With these changes, shouldn't you remove the `--watch` argument from the 
`CMD` line of the `frankenphp_dev` build traget ?

Then, without this argument, the `CMD` line would become the same than 
`frankenphp_base` and you can even remove it from `frakenphp_dev`, isn't 
it ?

dunglas:
`--watch` is something else, it hints Caddy to watch for changes in `Caddyfile`

maxhelias:
@dunglas Just to be sure, we now activate worker mode regardless of the environment and enable the watcher only in dev ?

ArmandArthur:
Where run --watch command?

PR #766, opened 2025-03-27
Update Mercure configuration in Caddyfile

7-zete-7:
Due to https://github.com/dunglas/mercure/pull/939 and the deprecation message, it makes sense to apply the Mercure configuration changes.

See:
- [Upgrade notice](https://github.com/dunglas/mercure/blob/main/docs/UPGRADE.md#017)
- [Original patch](https://github.com/dunglas/mercure/pull/939/commits/114044ee0fb80ebcb89c9f1f435a7f9fc422b911#diff-1f38639fad43624b85cdc2cac1778918713a061084bca6487683ec73f2b29fea)

### DIFF
--- a/frankenphp/Caddyfile
+++ b/frankenphp/Caddyfile
@@ -23,8 +23,6 @@
 	encode zstd br gzip
 
 	mercure {
-		# Transport to use (default to Bolt)
-		transport_url {$MERCURE_TRANSPORT_URL:bolt:///data/mercure.db}
 		# Publisher JWT key
 		publisher_jwt {env.MERCURE_PUBLISHER_JWT_KEY} {env.MERCURE_PUBLISHER_JWT_ALG}
 		# Subscriber JWT key


7-zete-7:
It might be more correct to replace it with the new configuration format

```diff
 {$SERVER_NAME:localhost} {
 	# ...
 	mercure {
+ 		# Transport to use (default to Bolt)
+		transport bolt {
+			path /data/mercure.db
+		}
```

However, this is already the default value. Backward compatibility (configuration via the environment variable `MERCURE_TRANSPORT_URL`) will also be broken.

dunglas (approved):

dunglas:
Simplifying the config is ok. `MERCURE_TRANSPORT_URL` BC is handled directly by Mercure.

7-zete-7:
I was too hasty with the default value. I found out that Mercure's default `path` option is `./bolt.db`. To keep the value the same, the `MERCURE_TRANSPORT_URL` environment variable must be explicitly set.

See https://github.com/dunglas/mercure/blob/6e73d7278ceb43c7cfd941089fe7bf984617a515/bolt.go#L88-L90

I'll add this as a suggestion to https://github.com/dunglas/symfony-docker/pull/767.

PR #767, opened 2025-03-27
docs: remove transport url config

maxhelias:
Fix a small oversight #766


7-zete-7:
There is no need to remove the configuration of this option. I was hasty in defining the default value for the Mercure path option 😅

Mercure will store its data at `./bolt.db` (`/app/bolt.db`) if the environment variable `MERCURE_TRANSPORT_URL` is not set. In https://github.com/dunglas/symfony-docker/pull/768, the use of this environment variable should return.

Issue #604, opened 2024-04-11
Using custom TLS certificates does not work


julien-guillon:
Hello,

I followed this part of the documentation on a **fresh install** : [Using Custom TLS Certificates](https://github.com/dunglas/symfony-docker/blob/main/docs/tls.md#using-custom-tls-certificates).

This error is displayed in the php container's logs :
```
Error: adapting config using caddyfile: server listening on [:80] is HTTP, but attempts to configure TLS connection policies
```

I even tried to start up by prefixing the command :
```
SERVER_NAME=server-name.localhost \
docker compose up --pull always -d --wait
```
or
```
SERVER_NAME=server-name.localhost:443 \
docker compose up --pull always -d --wait
```

Thanks in advance for your help!

YannickDurden:
Have you tried to edit composer.yaml file ?

`SERVER_NAME: ${SERVER_NAME:-localhost}, php:80 #use 443 instead of 80`

julien-guillon:
I did and Caddy is trying to obtain a certificate for "php".

```
ERROR tls.obtain will retry {"error": "[php] Obtain: [php] creating new order: attempt 1: https://acme.zerossl.com/v2/DV90/newOrder: HTTP 400 urn:ietf:params:acme:error:rejectedIdentifier - DNS identifier is invalid [php] (ca=https://acme.zerossl.com/v2/DV90)", "attempt": 1, "retrying_in": 60, "elapsed": 3.192137954, "max_duration": 2592000}
```

zerowebcorp:
This worked for me.

``` 
SERVER_NAME: ${SERVER_NAME:-dev.mywebsite.com}, app:80
```

Note: `app` is my service name in docker-compose file. 
It generated a public certificate using Google Trust

Bisb:
Try disabling Caddy [automatic https](https://caddyserver.com/docs/caddyfile/options#auto-https) in your `./compose.override.yaml`

```yaml
php:
  environment:
    CADDY_SERVER_EXTRA_DIRECTIVES: "tls /etc/caddy/certs/tls.pem /etc/caddy/certs/tls.key"
    CADDY_GLOBAL_OPTIONS: "auto_https off"
```

Perhaps this could make it into the Custom TLS doc

mcjwsk:
@Bisb it doesn't work:
```
php-1  | 2024/08/27 06:03:55.302        INFO    using config from file  {"file": "/etc/caddy/Caddyfile"}
[...]

PR #774, opened 2025-04-08
doc: update custom TLS certificate usage patch

7-zete-7:
This will allow you to configure a custom certificate and not get the following error:

```
Error: adapting config using caddyfile: server listening on [:80] is HTTP, but attempts to configure TLS connection policies
```

This issue was mentioned in:
- https://github.com/dunglas/symfony-docker/discussions/359
- https://github.com/dunglas/symfony-docker/discussions/579
- https://github.com/dunglas/symfony-docker/issues/604
- https://github.com/dunglas/symfony-docker/pull/738

With this configuration, Caddy will use the configured certificate only for access via a secure protocol.


7-zete-7:
I could not find information on how Caddy merges configurations of several addresses. I checked it on my projects and it works. If someone attaches links to information on how Caddy does this, I would be grateful.

docs/tls.md
```diff
@@ -30,11 +30,14 @@ For instance, to use self-signed certificates created with [mkcert](https://gith
...
       environment:
-    +    CADDY_SERVER_EXTRA_DIRECTIVES: "tls /etc/caddy/certs/tls.pem /etc/caddy/certs/tls.key"
+    +    CADDY_EXTRA_CONFIG: |
+    +      https:// {
+    +        tls /etc/caddy/certs/tls.pem /etc/caddy/certs/tls.key
+    +      }
         # ...
       volumes:
     +    - ./frankenphp/certs:/etc/caddy/certs:ro
-        - ./public:/app/public:ro
```
maxhelias, line 37: Do we need this change?

maxhelias:
> I could not find information on how Caddy merges configurations of several addresses. I checked it on my projects and it works. If someone attaches links to information on how Caddy does this, I would be grateful.

Maybe @dunglas or @mholt have the answer ?

mholt:
> Maybe @dunglas or @mholt have the answer ?

Sorry, I've looked at the linked issues but I'm not sure I understand what the question is exactly ("how Caddy merges configurations of several addresses"?). Is there a Caddy config in particular that is yielding that error?

The error means that the resulting Caddy config -- whatever it is -- has an HTTP server configured to listen on port 80, the HTTP port, but the server configures TLS connection policies, which enables TLS, making it HTTPS.

mholt:
@7-zete-7 Weird, I cannot reproduce the error with that config. It adapts to this JSON and seems to work fine:

```json
{
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":443"
          ],
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "localhost"
                  ]
                }
              ],
              "terminal": true
            }
          ],
          "tls_connection_policies": [
            {
              "certificate_selection": {
                "any_tag": [
                  "cert0"
                ]
              }
            }
          ]
        },
        "srv1": {
          "listen": [
            ":80"
          ],
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "php"
                  ]
                }
              ],
              "terminal": true
            }
          ]
        }
      }
    },
    "tls": {
      "certificates": {
        "load_files": [
          {
            "certificate": "/etc/caddy/certs/tls.pem",
            "key": "/etc/caddy/certs/tls.key",
            "tags": [
              "cert0"
            ]
          }
        ]
      }
    }
  }
}
```

Since only the `:443` server has TLS connection policies.

7-zete-7:
@mholt, I probably didn't communicate the information well, sorry about that 😅. This PR proposes a solution to the error, not its existence.

A configuration like https://github.com/dunglas/symfony-docker/pull/774#issuecomment-2790183619 solves the error.

I wanted to know information about how Caddy works when a site in a Caddyfile corresponds to several site blocks. I am curious myself, and I would attach this information to this PR to explain the proposed configuration.

I could only find the following information:

> To catch all hosts, omit the host portion of the address, for example, simply `https://`. This is useful when using [On-Demand TLS](https://caddyserver.com/docs/automatic-https#on-demand-tls), when you don't know the domains ahead of time.

Source: https://caddyserver.com/docs/caddyfile/concepts#addresses

This solution was based on this information. However, this information does not make it clear that the configurations will be merged, and this puzzles me.

maxhelias:
Oh yes, of course, now I remember — the initial problem isn’t really about the documentation, but rather that the user must specify the https scheme in their domain name. Otherwise (I think, not 100% sure), setting auto_https: disable_certs should also work.

Here it works because you're asking all HTTPS domains to use the custom TLS, see : https://caddyserver.com/docs/caddyfile/concepts#addresses

So maybe we should just add a note to explain that — what do you think?

mholt:
@7-zete-7 Ah, gotcha. Well, kind of:

> this information does not make it clear that the configurations will be merged

What do you mean by "the configurations will be merged"? The Caddyfile is adapted into JSON, it's a whole new config document. It's not really "merging" Caddyfile and JSON.

> how Caddy works when a site in a Caddyfile corresponds to several site blocks.

A site block does not correspond with other/several site blocks, except in this sense: A site block with a more specific site address will take priority over those with less specific matching site addresses. There is no inheritance or cascading, if that's what you mean?

@maxhelias I'm still not 100% sure I follow, but, if a site block defined as:

```
localhost, php:80 {
    # ....
}
```

has a `tls` directive in it, that will cause problems because you're telling Caddy to serve HTTPS on port 80.

7-zete-7:
@mholt, you are right. The words _inheritance_ or _cascading_ better fit what I was trying to say. Thanks for the correction.

Today I tried to play with the directives in Caddyfile and see how they affect the configuration in JSON (I used [Caddapto](https://caddapto.vercel.app/) for this). As a result, I got two directives that affect only the server, without creating a route. These are `log` and `tls`. And this seems quite logical. Working with the JSON configuration, this is even obvious. However, I could not find any mention of such behavior of these directives in the documentation. Although, as it turned out, a good indicator was the absence of a directive in the [Directive order](https://caddyserver.com/docs/caddyfile/directives#directive-order) section.

As a result, I got a Caddyfile where the site blocks, roughly speaking, extend each other:

<details>

<summary>The resulting Caddyfile</summary>

```Caddyfile
# creates apps.http.servers.srv0 and apps.http.servers.srv1 items
localhost, awesome-project.local, private.local:80 {
    respond "Hello, World!"
}

# doesn't not creates a new apps.http.servers item
# directives are relates to the localhost, awesome-project.local
https://, :443 {
    log {
        output stderr
    }

    tls public.pem private.pem
}

# doesn't not creates a new apps.http.servers item
# directives are relates to the private.local:80
http://, :80 {
    log {
        output stderr
    }
}
```

<details>

<summary>Configuration in JSON</summary>

```json
{
  "logging": {
    "logs": {
      "default": {
        "exclude": [
          "http.log.access.log0",
          "http.log.access.log1"
        ]
      },
      "log0": {
        "writer": {
          "output": "stderr"
        },
        "include": [
          "http.log.access.log0"
        ]
      },
      "log1": {
        "writer": {
          "output": "stderr"
        },
        "include": [
          "http.log.access.log1"
        ]
      }
    }
  },
  "apps": {
    "http": {
      "servers": {
        "srv0": {
          "listen": [
            ":443"
          ],
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "localhost",
                    "awesome-project.local"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "Hello, World!",
                          "handler": "static_response"
                        }
                      ]
                    }
                  ]
                }
              ],
              "terminal": true
            }
          ],
          "tls_connection_policies": [
            {
              "certificate_selection": {
                "any_tag": [
                  "cert0"
                ]
              }
            }
          ],
          "logs": {
            "default_logger_name": "log0"
          }
        },
        "srv1": {
          "listen": [
            ":80"
          ],
          "routes": [
            {
              "match": [
                {
                  "host": [
                    "private.local"
                  ]
                }
              ],
              "handle": [
                {
                  "handler": "subroute",
                  "routes": [
                    {
                      "handle": [
                        {
                          "body": "Hello, World!",
                          "handler": "static_response"
                        }
                      ]
                    }
                  ]
                }
              ],
              "terminal": true
            }
          ],
          "logs": {
            "default_logger_name": "log1"
          }
        }
      }
    },
    "tls": {
      "certificates": {
        "load_files": [
          {
            "certificate": "public.pem",
            "key": "private.pem",
            "tags": [
              "cert0"
            ]
          }
        ]
      }
    }
  }
}
```

</details>

</details>

Found a similar question in https://github.com/caddyserver/website/issues/235. I'll move the discussion on this behavior there.

maxhelias:
> I'm still not 100% sure I follow, but, if a site block defined as:
localhost, php:80 {
    # ....
}
has a tls directive in it, that will cause problems because you're telling Caddy to serve HTTPS on port 80.

That's what we do, so I guess that's where the problem is. 
Two suggest to fix your issue : 
- Recommend that the user use `auto_https: disable_certs` (not 100% sure) if they add a `tls` directive.
- Move that common part into a snippet, split the site blocks and import it in each block — makes it customizable and prevents the error.

7-zete-7:
@maxhelias, if I understand correctly, you are proposing a configuration of the following:

<details>

<summary>Suggested Configuration</summary>

```Caddyfile
# Caddyfile

{
    {$CADDY_GLOBAL_OPTIONS}

    # ...
}

(app) {
    # ...
}

{$CADDY_EXTRA_CONFIG}

{$SERVER_NAME:localhost} {
    import app

    ${CADDY_APP_SERVER_EXTRA_DIRECTIVES}
}

php:80 {
    import app

    ${CADDY_PRIVATE_SERVER_EXTRA_DIRECTIVES}
}
```

```diff yaml
 # compose.yaml

 services:
   php:
     # ...
     environment:
-      SERVER_NAME: ${SERVER_NAME:-localhost}, php:80
+      SERVER_NAME: ${SERVER_NAME:-localhost}
       # ...
+      CADDY_GLOBAL_OPTIONS: auto_https disable_certs
+      CADDY_APP_SERVER_EXTRA_DIRECTIVES: tls public.pem private.pem
       # ...
```

</details>

IMO Caddyfile has become a bit dirty. But it should definitely work. Although I'm not sure that `auto_https: disable_certs` will somehow change the behavior of Caddy for this configuration.

Perhaps we should stop using `php:80` and access the server by its `SERVER_NAME`? To prevent requests from going beyond the Docker network, add `SERVER_NAME` to [`service.php.network.aliases`](https://docs.docker.com/reference/compose-file/services/#aliases). But then the Mercure publisher will start accessing via HTTPS instead of HTTP. This may worsen performance. And the appearance of the `networks` attribute in the `php` service complicates this file.

<details>

<summary>Suggested Configuration</summary>

```diff yaml
 # compose.yaml

 services:
   php:
     # ...
+    networks:
+      default:
+        aliases:
+          - ${SERVER_NAME:-localhost}
     # ...
     environment:
-      SERVER_NAME: ${SERVER_NAME:-localhost}, php:80
+      SERVER_NAME: ${SERVER_NAME:-localhost}
     # ...
-    MERCURE_URL: ${CADDY_MERCURE_URL:-http://php/.well-known/mercure}
+    MERCURE_URL: ${CADDY_MERCURE_URL:-https://${SERVER_NAME:-localhost}/.well-known/mercure}
```

</details>

maxhelias:
No need to have `CADDY_GLOBAL_OPTIONS: auto_https disable_certs` with the snippet solution. It's either the snippet or the `CADDY_GLOBAL_OPTIONS: auto_https disable_certs` not both, but i'm not sure that the last solution fix the erreur.

mholt:
@7-zete-7 To clarify, is there anything wrong with that? The adapted config looks correct at a glance.

7-zete-7:
@maxhelias, it looks like I misunderstood the `CADDY_GLOBAL_OPTIONS: auto_https disable_certs` solution.

docs/tls.md
```diff
@@ -30,7 +30,10 @@ For instance, to use self-signed certificates created with [mkcert](https://gith
     ```diff
     php:
       environment:
-    +    CADDY_SERVER_EXTRA_DIRECTIVES: "tls /etc/caddy/certs/tls.pem /etc/caddy/certs/tls.key"
+    +    CADDY_EXTRA_CONFIG: |
+    +      https:// {
+    +          tls /etc/caddy/certs/tls.pem /etc/caddy/certs/tls.key
+    +      }
```
7-zete-7, line 36: There is incorrect formatting here (Caddyfile uses tabs for indentation)

```suggestion
    +      https:// {
    +      	tls /etc/caddy/certs/tls.pem /etc/caddy/certs/tls.key
    +      }
```

IMO it looks wrong visually (as if a single space is used for indentation, not a tab).

7-zete-7: In suggestion it looks better than it renders in Markdown
image

abdounikarim:
Hey,

I already has this error. I solve it this way.
Follow the docs here: https://github.com/dunglas/symfony-docker/blob/main/docs/tls.md#using-custom-tls-certificates
And I just updated this line in my `compose.yaml`: `SERVER_NAME: ${SERVER_NAME:-localhost}`

I didn't have to update anything in the `Caddyfile`.

WDYT ? 🤔

7-zete-7:
Hi, @abdounikarim!

I think you're right, removing the `php:80` host solves the problem, but it's not entirely correct for the current configuration.

You can disable the `php:80` host if you don't need Mercure (it's enabled by default). This host is used by the `MERCURE_URL` environment variable to access FrankenPHP from within the container.

abdounikarim:
Hi @7-zete-7 , you are right, I proposed this solution for people who doesn't use Mercure. 😀

7-zete-7:
Good news. It looks like the `MERCURE_URL` env and `php:80` may soon be unnecessary after a https://github.com/php/frankenphp/pull/1927. Issues with the `tls` directive should no longer occur by using actual documentation.

@abdounikarim, I'm guessing your solution become more correct and will soon be the default. And it won't even need to disable Mercure to do so.

abdounikarim:
That will be wonderful.
It will simplify the configuration with custom certificates.

I can update the doc about this if needed. 😉

Issue #776, opened 2025-04-09
Update the documentation pages of this template


7-zete-7:
Some of the documentation blocks in this template were written some time ago. Since then, there have been changes to the things mentioned in the documentation.

Some of the documentation pages also use mixed code styles. The pages need to be brought into the same format.

PR #778, opened 2025-04-09
chore: normalize docker compose commands

7-zete-7:
Part of #776

This PR brings the following commands to uniform formats:

* `docker compose ... build --pull --no-cache`

    * The options are set in this order (this order was more common in this template).

* `docker compose ... up --wait`

    * Specifying the `-d` option is optional (the `--wait` option implies the `-d` option, see https://github.com/dunglas/symfony-docker/issues/436#issuecomment-2792914977).

* `docker compose ... down --remove-orphans`

    Already in a unified format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ with [FrankenPHP](https://frankenphp.dev) and [Caddy](https://caddyserver.com/) 
 ## Getting Started
 
 1. If not already done, [install Docker Compose](https://docs.docker.com/compose/install/) (v2.10+)
-2. Run `docker compose build --no-cache` to build fresh images
-3. Run `docker compose up --pull always -d --wait` to set up and start a fresh Symfony project
+2. Run `docker compose build --pull --no-cache` to build fresh images
+3. Run `docker compose up --wait` to set up and start a fresh Symfony project
 4. Open `https://localhost` in your favorite web browser and [accept the auto-generated TLS certificate](https://stackoverflow.com/a/15076602/1352334)
 5. Run `docker compose down --remove-orphans` to stop the Docker containers.
 
--- a/docs/tls.md
+++ b/docs/tls.md
@@ -45,7 +45,7 @@ To disable HTTPS, configure your environment to use HTTP by setting the followin
 ```bash
 SERVER_NAME=http://localhost \
 MERCURE_PUBLIC_URL=http://localhost/.well-known/mercure \
-docker compose up --pull always -d --wait
+docker compose up --wait
 ```
 
 Ensure your application is accessible over HTTP by visiting `http://localhost` in your web browser.
--- a/frankenphp/docker-entrypoint.sh
+++ b/frankenphp/docker-entrypoint.sh
@@ -17,7 +17,7 @@ if [ "$1" = 'frankenphp' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		composer config --json extra.symfony.docker 'true'
 
 		if grep -q ^DATABASE_URL= .env; then
-			echo 'To finish the installation please press Ctrl+C to stop Docker Compose and run: docker compose up --build -d --wait'
+			echo 'To finish the installation please press Ctrl+C to stop Docker Compose and run: docker compose up --build --wait'
 			sleep infinity
 		fi
 	fi
--- a/docs/existing-project.md
+++ b/docs/existing-project.md
@@ -29,11 +29,11 @@ Double-check the changes, revert the changes that you don't want to keep:
 
 Build the Docker images:
 
-    docker compose build --no-cache --pull
+    docker compose build --pull --no-cache
 
 Start the project!
 
-    docker compose up -d
+    docker compose up --wait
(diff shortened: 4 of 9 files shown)


7-zete-7:
Command formats are based on the most popular variants in this template. Suggestions for better format options are welcome 👍

lermontex (commented):
It looks like this is another PR of yours that doesn't solve any issue.

The `--wait` flag must be used with detached mode.

> Wait for services to be running|healthy. **Implies detached mode.**

https://docs.docker.com/reference/cli/docker/compose/up/

Related Issue:  #436

> This PR brings the following commands to uniform formats

The commands are already applied in a uniform style (the long flag is applied in the absence of the short one)

Also, it seems like you have no idea what you're doing when you remove/add flags (`pull`, `always`, etc.) in the docs.

7-zete-7:
Thanks for the info, @lermontex!

I'm glad to get your critique!

> It looks like this is another PR of yours that doesn't solve any issue.
> 
> The `--wait` flag must be used with detached mode.
> 
> > Wait for services to be running|healthy. **Implies detached mode.**
> 
> https://docs.docker.com/reference/cli/docker/compose/up/
> 
> Related Issue: #436 

I haven't encountered this problem before and I'm curious to know what could be causing the problem. I didn't pay attention to this issue, my mistake. I'll probably change the unified format for this command.

I couldn't find an issue in the repository https://github.com/docker/compose about this problem. Maybe you have a link to something similar?

A very strange error. By the Docker Compose source code, the `-d` option is indeed explicitly enabled when the `--wait` option is present (see https://github.com/docker/compose/blame/01e83defc2e7c59c765be89509ce0f736dfcfc43/cmd/compose/up.go#L195-L200).

<hr>

> > This PR brings the following commands to uniform formats
> 
> The commands are already applied in a uniform style (the long flag is applied in the absence of the short one)

I can't agree with your statement. If you look at least in different parts of the documentation, you can see the presence and absence of the `--wait` option for the `docker compose up` command. Although they are needed in all cases.

<hr>

> Also, it seems like you have no idea what you're doing when you remove/add flags (`pull`, `always`, etc.) in the docs.

I have information from the documentation (see https://docs.docker.com/reference/cli/docker/compose/up/#options, https://docs.docker.com/reference/cli/docker/compose/build/#options), as well as personal experience in working with these options.

In my opinion, these options are used correctly in these changes.

If you have information on how this will change the behavior of the changed commands in a negative way, I would be grateful to receive information and make a rollback/corrections.

7-zete-7:
Converted to Draft until the issues that arose are resolved.

README.md
```diff
@@ -8,8 +8,8 @@ with [FrankenPHP](https://frankenphp.dev) and [Caddy](https://caddyserver.com/)
 ## Getting Started
 
 1. If not already done, [install Docker Compose](https://docs.docker.com/compose/install/) (v2.10+)
-2. Run `docker compose build --no-cache` to build fresh images
-3. Run `docker compose up --pull always -d --wait` to set up and start a fresh Symfony project
+2. Run `docker compose build --pull --no-cache` to build fresh images
+3. Run `docker compose up --wait` to set up and start a fresh Symfony project
```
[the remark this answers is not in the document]
7-zete-7: This command is left over from PHP-FPM + Caddy (see b2d0b4a21da5465c10b998aad077fb4d2faab746). Then pulling the `caddy` image was necessary. Now tamplate is designed to work with the container being built. There is no need to pull anything in the first steps.

docs/tls.md
```diff
@@ -45,7 +45,7 @@ To disable HTTPS, configure your environment to use HTTP by setting the followin
 ```bash
 SERVER_NAME=http://localhost \
 MERCURE_PUBLIC_URL=http://localhost/.well-known/mercure \
-docker compose up --pull always -d --wait
+docker compose up --wait
```
7-zete-7, line 48: Pulling the image was not required here. At the time of these steps, the images should be both built/pulled. I can assume that this was done in order to maintain a uniform format for the `docker compose up` command.

frankenphp/docker-entrypoint.sh
```diff
@@ -17,7 +17,7 @@ if [ "$1" = 'frankenphp' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 		composer config --json extra.symfony.docker 'true'
 
 		if grep -q ^DATABASE_URL= .env; then
-			echo 'To finish the installation please press Ctrl+C to stop Docker Compose and run: docker compose up --build -d --wait'
+			echo 'To finish the installation please press Ctrl+C to stop Docker Compose and run: docker compose up --build --wait'
```
7-zete-7, line 20: The `--build` option to the `docker compose up` command makes sense in this case. The Dockerfile can be changed after Symfony Flex actions.

7-zete-7:
The issues have been resolved at this time. Additional comments are welcome.

dunglas (approved):

maxhelias (approved):

PR #796, opened 2025-05-27
chore: Prevent "certbot is not available" warning by Caddy

### LINKED ISSUE #764: Missing certutil package
Hi there,

I just started a fresh new project.
Simply clone the repo and run `docker compose up -d`.

And then I noticed this in the first logs:
```
INFO    warning: "certutil" is not available, install "certutil" with "apt install libnss3-tools"  or "yum install nss-tools" and try again
```

Shouldn't we had this package in the Dockerfile ? 🤷 
This is not a blocking issue, the app is running without this anyway, so maybe we actually don't want to make the image bigger for 'no reason' ?

7-zete-7:
By default, Caddy expects to run in the operating system, not in a container. Because of this, Caddy's default behavior is to install its certificate as trusted (see https://caddyserver.com/docs/automatic-https#ca-root). This makes no sense in a container. However, Caddy will notify you if it fails to install the certificate as trusted with the following warning:

```plaintext
INFO    warning: "certutil" is not available, install "certutil" with "apt install libnss3-tools"  or "yum install nss-tools" and try again
```

Since [v2.4.2](https://github.com/caddyserver/caddy/releases/tag/v2.4.2), Caddy has a global option `skip_install_trust` that disables Caddy's attempts to install its certificate as trusted. Having this global option disables this warning and may slightly improve container startup speed.

Fixes #764


dunglas (approved):
You can even remove the comment IMHO

francislavoie:
> This makes no sense in a container

That's not quite true. Caddy needs to have its own cert in the trust store if you're doing something like having Caddy connect to itself (e.g. `reverse_proxy https://localhost`, which is a valid usecase in certain situations.

7-zete-7:
Thanks for pointing this out, @francislavoie!

I've made this statement more specific in the PR description.

For context, I'll attach a link to the HTTPS testing proposed in this template below. It uses the `-k`/`--insecure` cURL option. This template does not require trusting Caddy's own certificate for correct operation.

https://github.com/dunglas/symfony-docker/blob/46e07dfce7f06865ba83069fd330b4d111914fc4/.github/workflows/ci.yml#L44-L47

francislavoie:
`-k` basically turns off all security, meaning using HTTPS has _no advantage at all_ over HTTP because it can be trivially man-in-the-middle'd. IMO, either default to HTTP for dev, or document clearly how to install the cert (covered here in Caddy docs https://caddyserver.com/docs/running#local-https-with-docker)

7-zete-7:
@francislavoie , I agree with you. IMO the `-k` option should not be used if there are other solutions (double-checked). By using the `-k` option, I meant that trusting the Caddy certificate is not required by default in this template.

Also this repository already has a document describing several ways to fix the HTTPS issue (see [TLS Certificates](https://github.com/dunglas/symfony-docker/blob/15dbd50eee5e84aa4bd807f7256260ad407eb72f/docs/tls.md)). A method similar to the one proposed in the Caddy documentation is suggested first.

AFAIU this template does not plan to install the `libnss3-tools` package. Because of this, the check for the `certutil` utility will always fail with a warning. This PR essentially disables the check and hides the warning.

With the global `skip_install_trust` option, Caddy's internal certificates are still generated and used. If necessary, all of the methods listed in the [TLS Certificates](https://github.com/dunglas/symfony-docker/blob/15dbd50eee5e84aa4bd807f7256260ad407eb72f/docs/tls.md) document remain functional for the user of this template.

The Caddyfile of this template is intended to be used only inside containers and only in the environment generated by this template. In my opinion (and a couple of tests), the presence of the global `skip_install_trust` option for this repository does not change the way the resulting project interacts in any way. I would also not recommend using this global option outside the context of this template.

francislavoie:
I would recommend not using `-k` and rather using `--insecure` to make it clear that it's bad to use. When you have `-vk` like that it's not obvious that there's a significant security downgrade happening.

7-zete-7:
@francislavoie, I think this is a good idea. Created a PR to implement this change: https://github.com/dunglas/symfony-docker/pull/799

PR #799, opened 2025-06-04
chore(ci): Make the security downgrade of cURL requests more obvious

7-zete-7:
Proposal of this change: https://github.com/dunglas/symfony-docker/pull/796#issuecomment-2939808993

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,10 +44,10 @@ jobs:
       -
         name: Check HTTPS reachability
         if: false # Remove this line when the homepage will be configured, or change the path to check
-        run: curl -vk --fail-with-body https://localhost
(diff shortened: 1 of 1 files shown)


dunglas (approved):

lermontex:
What's the point, considering that:

`"-k" === "--insecure"`

This is just an internal test that has nothing to do with security.

If this is for a more complete understanding, then why do you need to use the full form for selective parameters, and not for all?

Something like this:
`curl --verbose --insecure --fail-with-body https://localhost`
Instead of:
`curl -vk --fail-with-body https://localhost`

IMHO: Another pointless PR

maxhelias (approved):

maxhelias:
Thanks @7-zete-7 !

PR #803, opened 2025-06-13
Use COMPOSE_BAKE=true for better performance

abdounikarim:
Hey,

I just add a small improvment to speed up docker compose builds.


7-zete-7:
Hi @abdounikarim!

As of Docker Compose [v2.37.0](https://github.com/docker/compose/releases/tag/v2.37.0), this is no longer necessary. [Bake is used by default](https://github.com/docker/compose/pull/12699). In future releases, using this environment variable will result in deprecation messages.

PR #808, opened 2025-06-20
fix(doc): worker mode

94noni:
as far as I understand, this doc was for previous version of the project, now worker seem to be included always and not disabled by such env var


maxhelias (approved):

maxhelias:
Yes, good catch. Thanks

PR #823, opened 2025-08-04
feat: hide dot env files

jfcoz:
Hide `.env` and `.env.*` files


7-zete-7:
Hi @jfcoz!

This is not necessary because the .env files are in the `/app` directory, and the web server only provides access to files in the `/app/public` directory.

PR #824, opened 2025-08-13
Mark /app as safe directory for Git

danut007ro:
In my github actions I got error

```
The repository at "/app" does not have the correct ownership and git refuses to use it:

fatal: detected dubious ownership in repository at '/app'
To add an exception for this directory, call:

	git config --global --add safe.directory /app
```

This PR fixes the issue by adding an actions to mark /app directory as safe for git.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,9 @@ jobs:
         name: Start services
         run: docker compose up --wait --no-build
       -
+        name: Mark /app as safe directory for Git
+        run: docker compose exec -T php git config --global --add safe.directory /app
+      -
         name: Check HTTP reachability
         run: curl -v --fail-with-body http://localhost
       -
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,9 +39,6 @@ jobs:
         name: Start services
         run: docker compose up --wait --no-build
       -
-        name: Mark /app as safe directory for Git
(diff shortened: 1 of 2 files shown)


7-zete-7:
Hi @danut007ro!

I would suggest adding this command to the Dockerfile for the `frankenphp_dev` target. WDYT?

danut007ro:
Sure @7-zete-7 , have a look now. Thanks

7-zete-7 (approved):

maxhelias:
Hum... I don’t think we should add this.
Git isn’t meant to be used inside the container, and this just bypasses the “dubious ownership” warning instead of addressing a real need.
It also adds an unnecessary global config inside the image.

danut007ro:
Closing, as I am unable to reproduce on a new project. Will reopen if needed. Thank you.

danut007ro:
Managed to reproduce on a new clean project. Here's the PR on that new project https://github.com/danut007ro/symfony-docker-test/pull/1/files with result https://github.com/danut007ro/symfony-docker-test/actions/runs/16966099280/job/48090334555?pr=1

As you can see, the warning occurs when running `composer audit` command. Maybe there is another way to fix it?

Screenshot 2025-08-14 at 16 15 26

maxhelias:
As explain here : https://github.com/dunglas/symfony-docker/pull/824#issuecomment-3187638667 I'm not convinced about including that.
Thank you anyway.

PR #826, opened 2025-08-16
feat(docker): add Symfony CLI in php image

elmehdi-elmais:
- Install Symfony CLI only in the frankenphp_dev target
- Allows running `symfony console` commands inside the dev container
- Keeps the production image lighter since Symfony CLI is not required there

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,6 +60,9 @@ ENV APP_ENV=dev
 ENV XDEBUG_MODE=off
 ENV FRANKENPHP_WORKER_CONFIG=watch
 
+RUN curl -1sLf 'https://dl.cloudsmith.io/public/symfony/stable/setup.deb.sh' |  bash 
+RUN apt install symfony-cli
+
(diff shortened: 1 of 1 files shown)


maxhelias:
Symfony Docker provides the features of Symfony CLI. If you use Symfony Docker, you don't need Symfony CLI (and if you use Symfony CLI, you likely don't need/want Symfony Docker).

elmehdi-elmais:
Thanks for your feedback 
I know Symfony Docker already covers most of what Symfony CLI does, but I find it handy to have "symfony console" directly in the dev container.  

Since this is only added to "frankenphp_dev", it doesn’t affect production.  
For me it’s just a small convenience for devs at no real cost, but happy to adjust if you think otherwise.

dunglas:
This will confuse users, bloat the image and add nothing useful.

Sorry but we try to keep these images as minimal as possible.

elmehdi-elmais:
Got it, thanks for clarifying.

PR #828, opened 2025-08-19
RFC: Update Dockerfile for --classmap-authoritative

94noni:
RFC: is this ok and similar ?

leverage `--classmap-authoritative` with `install` not via independent `dump-autoload`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -83,7 +83,7 @@ COPY --link frankenphp/conf.d/20-app.prod.ini $PHP_INI_DIR/app.conf.d/
 # prevent the reinstallation of vendors at every changes in the source code
 COPY --link composer.* symfony.* ./
 RUN set -eux; \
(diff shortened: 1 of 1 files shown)


7-zete-7:
Hi @94noni!

I see two problems:
- The `--classmap-authoritative` option will be ignored due to the presence of the `--no-autoloader` option.
- Generating an autoloader at the time of executing the `composer install …` command of this Dockerfile will not dump the correct autoloader because the project files do not exist at this moment.

94noni:
@7-zete-7 hum ok I see very well with explanations thank you
is this also the reason why `--no-autoloader` is defined ? so the dump autoload do it later ?

7-zete-7:
@94noni yes, you got that right
IMO this is a clever optimization of the container build

94noni:
fair thx for answering !
lets close then :)

PR #829, opened 2025-08-19
deps: remove unnecessary dependencies

7-zete-7:
I suggest removing the following dependencies:
- `acl`
- `gettext`

These dependencies are not used or have no useful functionality.

See https://github.com/dunglas/symfony-docker/discussions/819

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,9 +18,7 @@ VOLUME /app/var/
 # persistent / runtime deps
 # hadolint ignore=DL3008
 RUN apt-get update && apt-get install -y --no-install-recommends \
-	acl \
 	file \
-	gettext \
 	git \
 	&& rm -rf /var/lib/apt/lists/*
 
--- a/docs/alpine.md
+++ b/docs/alpine.md
@@ -16,16 +16,12 @@ To switch to Alpine-based images, apply the following changes to the `Dockerfile
 
 -# hadolint ignore=DL3008
 -RUN apt-get update && apt-get install -y --no-install-recommends \
--	acl \
 -	file \
--	gettext \
 -	git \
 -	&& rm -rf /var/lib/apt/lists/*
 +# hadolint ignore=DL3018
 +RUN apk add --no-cache \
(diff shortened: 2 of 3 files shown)


dunglas:
IIRC, the file binary is used by Symfony (HttpFoundation?) to guess the mime type of uploaded files.

I'm not sure about the other tools. We used to use `acl` in the Docker entrypoint, but it may not be the case anymore.

7-zete-7:
@dunglas you are right. Found usage of the `file` binary in the [Symfony\Component\Mime\FileBinaryMimeTypeGuesser](https://github.com/symfony/symfony/blob/6ce530c5e90397d88e3a76a56db266c74d651584/src/Symfony/Component/Mime/FileBinaryMimeTypeGuesser.php#L33) class (migrated from HttpFondation). Will undo removal of that package.

Thanks for pointing this out!

danut007ro:
If `git` is removed then this should also solve #824

zpi12lmm:
Dependency `acl` is used here:
1. https://github.com/dunglas/symfony-docker/blob/main/frankenphp/docker-entrypoint.sh#L60
2. https://github.com/dunglas/symfony-docker/blob/main/frankenphp/docker-entrypoint.sh#L61

7-zete-7:
@danut007ro I agree. But I'm not sure if git can be a mandatory dependency for some Composer scenarios.

Maybe there is some information why git might be needed for Composer and whether this dependency is important?

7-zete-7:
@zpi12lmm yes, and this PR removes them.

These lines do not carry any useful payload within this template. The container always starts as root by default (and some entrypoint lines will only work as root), so there is no need to grant file permissions to www-data and the current user (the same root).

maxhelias (approved):

maxhelias:
@7-zete-7 Can you resolve conflicts? pls

7-zete-7:
@maxhelias Resolved conflicts

maxhelias:
Thanks!

PR #860, opened 2025-10-16
Fix Composer installation by using OpenSSL 3.0.17 in FrankenPHP base image

powernic:
**Description:**
The build of the Symfony Docker image fails at the step `install-php-extensions @composer` with the following error:

```
curl: (35) Recv failure: Connection reset by peer
```

This is caused by the OpenSSL version `3.5.1` included in `dunglas/frankenphp:1-php8.4`, which breaks HTTPS connections to `https://repo.packagist.org`.

To fix this, I propose switching the base image to a version of FrankenPHP that still uses OpenSSL `3.0.17`. This allows Composer to successfully connect and install dependencies over HTTPS.

**Changes:**

* Use `dunglas/frankenphp:1.9.0-php8.4` as the base image instead of the current one that ships OpenSSL 3.5.1.

**Impact:**

* Composer installation will work correctly during the Docker build.
* No changes to application code are required.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 #syntax=docker/dockerfile:1
 
 # Versions
-FROM dunglas/frankenphp:1-php8.4 AS frankenphp_upstream
+FROM dunglas/frankenphp:1.9.0-php8.4 AS frankenphp_upstream
 
 # The different stages of this Dockerfile are meant to be built into separate images
 # https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage


7-zete-7:
Hi, @powernic!

I've tested the `install-php-extensions @composer` step several times for the `dunglas/frankenphp:1-php8.4` image and haven't gotten the error you described.

I tested it with the following command:

```console
docker run --rm -it dunglas/frankenphp:1-php8.4 install-php-extensions @composer apcu intl opcache zip
```

I've tested it on different servers and on my local environment.

Can you describe in more detail the environment in which this error can be reproduced? Can you attach any additional information about this issue?

powernic:
Hi  @7-zete-7👋

Thank you for checking!
This issue occurs in a WSL environment (Windows Subsystem for Linux), where Docker is installed directly inside Ubuntu, not via Docker Desktop.

In this setup, the network connection to https://repo.packagist.org fails during the install-php-extensions @composer step when using openssl 3.5.1 (the version bundled in dunglas/frankenphp:1-php8.4).
The same step succeeds when using an image based on an older OpenSSL version (for example, 3.0.17).

It looks like OpenSSL 3.5.x introduced changes that cause TLS handshake resets in certain WSL network configurations.

7-zete-7:
Thanks for the details!

I'm guessing this is a very specific environment that's rarely encountered. I'm also guessing that FrankenPHP 1.9.0 images can also be updated to the new Debian. Therefore, I suspect this change may not resolve your issue.

I suggest adding the following line above `install-php-extensions @composer`:

```Dockerfile
RUN set -eux; \
	curl http://security.debian.org/debian-security/pool/updates/main/o/openssl/openssl_3.0.17-1~deb12u3_amd64.deb -o openssl.deb; \
	dpkg -i openssl.deb; \
	rm openssl.deb
```

This step will replace OpenSSL 3.5.1 with 3.0.17. This might help if the issue is indeed related to the OpenSSL version.

powernic:
Yes, indeed, your suggested approach solves the problem! Could you please consider adding this step to the Dockerfile? This would ensure that the `install-php-extensions @composer` step works reliably in environments like mine.

Thank you!

7-zete-7:
Another solution is to explicitly specify that Debian bookworm is expected. To do this, simply use the `dunglas/frankenphp:1-php8.4-bookworm` image.

PR #863, opened 2025-10-26
Add asset-map compilation if importmap is defined

florentdestremau:
This is useful for plug-and-play production mode for asset-mapper

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -92,4 +92,7 @@ RUN set -eux; \
 	composer dump-autoload --classmap-authoritative --no-dev; \
 	composer dump-env prod; \
 	composer run-script --no-dev post-install-cmd; \
+	if [ -f importmap.php ]; then \
+		php bin/console asset-map:compile; \
+	fi; \
 	chmod +x bin/console; sync;


frankenphp/docker-entrypoint.sh
```diff
@@ -56,6 +56,11 @@ if [ "$1" = 'frankenphp' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 			php bin/console doctrine:migrations:migrate --no-interaction --all-or-nothing
 		fi
 	fi
+	
+	if [ -f importmap.php ]; then
+		php bin/console asset-map:compile
+	fi
+
```
dunglas:
```suggestion
```

dunglas (approved):

bkosun (changes requested):
Please note that according to the AssetMapper documentation, asset compilation should be performed in the production environment, otherwise Symfony will stop serving assets dynamically:

> If you run the `asset-map:compile` command on your development machine, you won't see any changes made to your assets when reloading the page. To resolve this, delete the contents of the `public/assets/` directory. This will allow your Symfony application to serve those assets dynamically again.

https://symfony.com/doc/7.3/frontend/asset_mapper.html#serving-assets-in-dev-vs-prod

It is also necessary to provide for the possibility of installing dependencies, as they may not be installed:

> All packages in `importmap.php` are downloaded into an `assets/vendor/` directory, which should be ignored by git (the Flex recipe adds it to `.gitignore` for you). You'll need to run the following command to download the files on other computers if some are missing:
`php bin/console importmap:install`

https://symfony.com/doc/7.3/frontend/asset_mapper.html#importing-3rd-party-javascript-packages

In addition, the current solution will run dependency installation and asset compilation every time the container is started, and even if various checks are added, this is not the best solution, since building the project at this stage takes time (the result is not cached) and can lead to various errors (e.g., due to network issues). That is why the installation of Composer dependencies for the production environment is performed during image build:

https://github.com/dunglas/symfony-docker/blob/322b86959dbe028b16c90828646eed51867b8bec/Dockerfile#L82-L95

florentdestremau:
> Please note that according to the AssetMapper documentation, asset compilation should be performed in the production environment, otherwise Symfony will stop serving assets dynamically:

I added a check on APP_ENV so that this is only run in prod mode, you are right :)

As for using this in build phase instead of run phase, I tried adding this to the dockerfile in the frankenphp_prod build phase, at the very end of the file:

```
RUN if [ -f importmap.php ]; then \
    php bin/console importmap:install;\
    php bin/console asset-map:compile;\
fi
```

This works, and seems better placed then ?

7-zete-7:
The path to `importmap.php` can be configured differently using `asset_mapper.importmap_path`.

Does it make sense to replace the check for the existents of the `importmap.php` file with a command like `php bin/console list --quiet --no-interaction 'asset-map' 2>/dev/null`?

florentdestremau:
Launching a full command would be slower, another way would be to check composer.json graph ?
Like this:
```bash
    if composer show symfony/asset-mapper >/dev/null 2>&1; then \
        php bin/console asset-map:compile; \
    fi; \
```

Or even simpler with a grep ?

```bash
    if grep -q '"symfony/asset-mapper"' composer.json; then \
        php bin/console asset-map:compile; \
    fi; \
```

To be honest, I'm not sure a template should handle the weird use case where importmap is elsewhere. Looking at the conditions in the docker-entrypoint.sh, 	`if grep -q ^DATABASE_URL= .env; then` is already opiniated on the fact that you use full URL env var for your database, and not the DATABASE_PORT, DATABASE_PASSWORD, etc.

Same for migrations:  `if [ "$(find ./migrations -iname '*.php' -print -quit)" ]; then`

So if you stray from the default config you are already accepting that this template's defaults _might_ not work out

maxhelias:
Thanks @florentdestremau !

PR #865, opened 2025-10-29
Fix doctrine mysql deprecation

philbates35:
After following the docs to switch from Postgres to MySQL on a fresh project with `symfony/orm-pack` installed, (and `MYSQL_VERSION` env var is not defined meaning it falls back to the default version), I noticed this deprecation when running `bin/console doctrine:schema:validate -vvv`:
```
User Deprecated: Version detection logic for MySQL will change in DBAL 4. Please specify the version as the server reports it, e.g. "8.0.31" instead of "8". (AbstractMySQLDriver.php:94 called by AbstractDriverMiddleware.php:68, https://github.com/doctrine/dbal/pull/5779, package doctrine/orm)
```
The reason I went for `8.0.32` is because when the new project is scaffolded, the `.env` file specifies version 8.0.32 by default because of the doctrine-bundle recipe, so it makes sense that it would match.

https://github.com/symfony/recipes/blob/main/doctrine/doctrine-bundle/2.13/manifest.json#L14

```
"#5": "DATABASE_URL=\"mysql://app:!ChangeMe!@127.0.0.1:3306/app?serverVersion=8.0.32&charset=utf8mb4\"",
```

### DIFF
--- a/docs/mysql.md
+++ b/docs/mysql.md
@@ -20,7 +20,7 @@ Change the database image to use MySQL instead of PostgreSQL in `compose.yaml`:
 ```diff
  ###> doctrine/doctrine-bundle ###
 -    image: postgres:${POSTGRES_VERSION:-16}-alpine
-+    image: mysql:${MYSQL_VERSION:-8}
++    image: mysql:${MYSQL_VERSION:-8.0.32}
      environment:
 -      POSTGRES_DB: ${POSTGRES_DB:-app}
 +      MYSQL_DATABASE: ${MYSQL_DATABASE:-app}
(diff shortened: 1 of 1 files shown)


dunglas:
I would rather update the recipe. If we do that, we'll recommend installing outdated MySQL versions that may have known security vulnerabilities.

philbates35:
@dunglas Just to be clear I guess there's two things going on:

1. The reason I saw the deprecation is because I applied the diff from the readme which doesn't specify a full patch version, only `8`. The doctrine-bundle recipe does specify the full patch version in it's recipe, so there's nothing to change in the recipe to fix the deprecation which was the main purpose of the PR.
2. The recipe has an outdated version specified in its example `.env`. We could indeed look to update this in a PR to their repo, but it would need updating every time a patch or minor mysql release happened, which doesn't seem realistic and isn't related to the deprecation I was trying to fix.

    Also worth noting that the `DATABASE_URL` added to the `.env` by the recipe isn't even used in a symfony-docker project anyway because `DATABASE_URL` is also defined in the container env vars defined in this project which have a higher priority than env vars defined in `.env`, so making changes to the recipe doesn't actually have any impact at all on symfony-docker.

    We could just use the latest patch release in this PR if you prefer, it just means that when you install orm-pack the example value in the `.env` (which as I say isn't actually used) it would have an older version (8.0.32) than what the project is using (8.0.41 or whatever the latest is at the point the project is created)?

PR #866, opened 2025-11-03
added Makefile

des1roer:


dunglas:
Thanks for your contribution.

We don't want to provide a Makefile in the main repository. There is already a sample one provided in the docs. Feel free to improve it if needed.

Best regards,

PR #869, opened 2025-11-24
fix: remove unnecessary configuration due new `mercure_publish()` PHP function

7-zete-7:
With the addition of the new PHP function `mercure_publish()` in https://github.com/php/frankenphp/pull/1927, the use of the closure via the `php:80` host is no longer necessary. The `php:80` can be removed and the `MERCURE_URL` env can be ~~removed~~ empty.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -3,14 +3,14 @@ services:
   php:
     restart: unless-stopped
     environment:
-      SERVER_NAME: ${SERVER_NAME:-localhost}, php:80
+      SERVER_NAME: ${SERVER_NAME:-localhost}
       DEFAULT_URI: https://${SERVER_NAME:-localhost}:${HTTPS_PORT:-443}
       MERCURE_PUBLISHER_JWT_KEY: ${CADDY_MERCURE_JWT_SECRET:-!ChangeThisMercureHubJWTSecretKey!}
       MERCURE_SUBSCRIBER_JWT_KEY: ${CADDY_MERCURE_JWT_SECRET:-!ChangeThisMercureHubJWTSecretKey!}
       # Run "composer require symfony/orm-pack" to install and configure Doctrine ORM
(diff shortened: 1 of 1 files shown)


7-zete-7:
The [actual `symfony/mercure-bundle` recipe](https://github.com/symfony/recipes/blob/7e8de6e509e24e47bda885dcd227115b8e800159/symfony/mercure-bundle/0.3/config/packages/mercure.yaml) still requires the `MERCURE_URL` env. I'm guessing it's not yet possible to prevent using the environment variable in this recipe, as other environments expect this behavior.

7-zete-7:
Waiting until https://github.com/symfony/recipes/pull/1514 is merged.

dunglas:
@7-zete-7 can we merge this one?

7-zete-7:
@dunglas yes. I rebased it and double-checked it -- it works as expected.

7-zete-7:
@dunglas unfortunately, something went wrong. After installing the debug package, the container stopped building. I'll keep an eye on it. In the meantime, I'm converting it back to draft.

```
In Hub.php line 30:
                                                                                                                                                                 
  Symfony\Component\Mercure\Hub::__construct(): Argument #1 ($url) must be of type string, null given, called in /app/var/cache/dev/Container4tL1W4Y/App_Kernel  
  DevDebugContainer.php on line 799                                                                                                                                                                                                                                                                                            
```

7-zete-7:
Found the root cause of the issue: `MercureExtension` incorrectly evaluates the `$builtinHub` variable when `env(...)` is used in the `mercure.url` configuration. It generates a string value like `env_3b1ca2e0faf93766_default__MERCURE_URL_fef93def10a807a7253b2bc8bb994978`, which is interpreted as a populated `mercure.url`.

I'll create a patch for `MercureExtension`.

PR #871, opened 2025-11-26
fix(mercure)!: fix Mercure transport configuration

### LINKED ISSUE #870: Mercure transport is not configurable since FrankenPHP v1.10
Since Mercure v0.21, it's now possible to disable deprecated transport behavior (or rather, enable it with the `deprecated_transport` flag) (see https://github.com/dunglas/mercure/pull/1105). By default, Mercure is built with this flag.

FrankenPHP is built without the `deprecated_transport` flag. This causes the `MERCURE_TRANSPORT_URL` env to be ignored. The default transport is `bolt:///bolt.db`.

This behavior shouldn't affect most users. However, there are exceptions:

- Those who used the `MERCURE_TRANSPORT_URL` environment variable may find that its value has no effect.

- When running as an unprivileged user, the following container startup error will occur:

  ```
  Error: loading initial config: loading new config: loading frankenphp app module: provision frankenphp: failed to provision caddy http: loading http app module: provision http: server srv0: setting up route handlers: route 0: loading handler modules: position 0: loading module 'subroute': provision http.handlers.subroute: setting up subroutes: route 2: loading handler modules: position 2: loading module 'mercure': provision http.handlers.mercure: provision http.handlers.mercure.bolt: "": invalid transport: open bolt.db: permission denied
  ```

I see solutions to this problem:

- Stop using the `MERCURE_TRANSPORT_URL` env and start using the `mercure.transport` directive (see https://github.com/dunglas/mercure/blob/c7cf56e647e87aa814331e5ebf6672cae1ed2a54/docs/UPGRADE.md#017)
- Add the `deprecated_transport` flag when building FrankenPHP
- Make the deprecated Mercure transport behavior the default

I suggest the first option.

@dunglas @maxhelias

7-zete-7:
Fixes #870

The `MERCURE_TRANSPORT_URL` env has been removed and replaced by the `mercure.transport` Caddy directive.

BREAKING CHANGE: The `MERCURE_TRANSPORT_URL` env is no longer available.

### Update guide

Instead of using the `MERCURE_TRANSPORT_URL` env, use the `MERCURE_EXTRA_DIRECTIVES` env.

- Old usage:

  ```dotenv
  MERCURE_TRANSPORT_URL=bolt:///var/mercure.db
  ```

  New usage:

  ```yaml
  # compose.yaml
  services:
    php:
      # ...
      environment:
        # ...
        MERCURE_EXTRA_DIRECTIVES: |-
          transport bolt {
            path /var/mercure.db
          }
  ```

- Old usage:

  ```dotenv
  MERCURE_TRANSPORT_URL=bolt:///data/mercure.db?cleanup_frequency=0.2
  ```

  New usage:

  ```yaml
  # compose.yaml
  services:
    php:
      # ...
      environment:
        # ...
        MERCURE_EXTRA_DIRECTIVES: |-
          transport bolt {
            path /data/mercure.db
            cleanup_frequency 0.2
          }
  ```

See https://github.com/dunglas/mercure/blob/c7cf56e647e87aa814331e5ebf6672cae1ed2a54/docs/UPGRADE.md#017


shulard:
Hello !

I faced the same issue (having the bolt.db file at my project root). With this MR it's mandatory to use the bolt transport driver by default.

Do you think it's interesting to also add variable for the transport type like this:

```
transport {$MERCURE_TRANSPORT_TYPE:bolt} {
    path {$MERCURE_BOLT_PATH:/data/mercure.db}
    {$MERCURE_BOLT_EXTRA_DIRECTIVES}
}
```

Or maybe if we define another transport in the `MERCURE_EXTRA_DIRECTIVES` it'll override the previous definition ?

7-zete-7:
Hi, @shulard!

Adding the ability to configure transports via an environment variable like `MERCURE_TRANSPORT_TYPE` won't work because transports have different configuration options (see the Transports section at https://mercure.rocks/docs/hub/cluster).

Defining a different transport in the `MERCURE_EXTRA_DIRECTIVES` does indeed override the previous directive. But I'm not sure this is a valid configuration and won't cause errors in the future.

In some of my projects, I removed the entire `transport` directive from the Caddyfile and moved it to the `MERCURE_EXTRA_DIRECTIVES` environment variable in the compose.yaml file as the default behavior. This makes it possible to configure transports via, for example, the compose.prod.yaml file. However, this results in a multi-line environment variable and a more complex compose.yaml file.

shulard:
Ok thank you for the feedback. If it works in `MERCURE_EXTRA_DIRECTIVES` with multi line env var it's fine.

dunglas:
IMHO we should just set `MERCURE_EXTRA_DIRECTIVES` as well as a volume to store the bolt file in `compose.yaml`

shulard:
> as well as a volume to store the bolt file in compose.yaml

Do you think it's really important to have a specific volume ? If we put the db file in `/data` it's already covered by the `caddy_data` volume.

dunglas:
You're right, data is good enough.

dunglas:
WDYT of https://github.com/dunglas/mercure/pull/1139?

This should entirely remove this problem.

shulard:
> WDYT of [dunglas/mercure#1139](https://github.com/dunglas/mercure/pull/1139)?
> 
> This should entirely remove this problem.

This would be nice since it'll not require any configuration in that repository :smile:.

7-zete-7:
> WDYT of https://github.com/dunglas/mercure/pull/1139?

>
> This should entirely remove this problem.

This looks great to me!

I'll leave this PR in its current state for now so that anyone who encounters the same issue can take advantage of this fix.
Once FrankenPHP is released with the https://github.com/dunglas/mercure/pull/1139, I'll remove the transport configuration here.

dunglas:
I tagged the new release. Removing old directives should be enough now.

tcoch:
Hi everyone,
Isn't it OK to merge this PR now ?

Fuzip (approved):

dunglas (approved):

dunglas:
Thanks

PR #873, opened 2025-11-30
fix(deps)!: Remove `runtime/frankenphp-symfony` dependency

7-zete-7:
The latest version of Symfony v7.4 introduces native runner support for FrankenPHP (see https://github.com/symfony/symfony/pull/60503). The current new project initialization script explicitly adds the `runtime/frankenphp-symfony` dependency, which is no longer needed for projects running the latest version of Symfony.

For Symfony v8.0 (the default version), adding the `runtime/frankenphp-symfony` package results in an error (see https://github.com/dunglas/symfony-docker/issues/872).

<details><summary>Details</summary>

```plain
Creating a "symfony/skeleton " project at "./tmp"
The repository at "/app" does not have the correct ownership and git refuses to use it:

fatal: detected dubious ownership in repository at '/app'
To add an exception for this directory, call:

       git config --global --add safe.directory /app

Installing symfony/skeleton (v8.0.99)
  - Downloading symfony/skeleton (v8.0.99)
  - Installing symfony/skeleton (v8.0.99): Extracting archive
Created project in /app/tmp
/app
The repository at "/app" does not have the correct ownership and git refuses to use it:

fatal: detected dubious ownership in repository at '/app'
To add an exception for this directory, call:

       git config --global --add safe.directory /app

./composer.json has been updated
The repository at "/app" does not have the correct ownership and git refuses to use it:

fatal: detected dubious ownership in repository at '/app'
To add an exception for this directory, call:

       git config --global --add safe.directory /app

Running composer update php runtime/frankenphp-symfony
Loading composer repositories with package information
Updating dependencies
Lock file operations: 18 installs, 0 updates, 0 removals
  - Locking psr/container (2.0.2)
  - Locking psr/event-dispatcher (1.0.0)
  - Locking psr/log (3.0.2)
  - Locking runtime/frankenphp-symfony (0.2.0)
  - Locking symfony/dependency-injection (v7.4.0)
  - Locking symfony/deprecation-contracts (v3.6.0)
  - Locking symfony/error-handler (v8.0.0)
  - Locking symfony/event-dispatcher (v8.0.0)
  - Locking symfony/event-dispatcher-contracts (v3.6.0)
  - Locking symfony/flex (v2.10.0)
  - Locking symfony/http-foundation (v8.0.0)
  - Locking symfony/http-kernel (v7.4.0)
  - Locking symfony/polyfill-mbstring (v1.33.0)
  - Locking symfony/polyfill-php85 (v1.33.0)
[...]

PR #874, opened 2025-11-30
fix(lint): Ignore always-false condition errors in ci workflow

7-zete-7:
Since actionlint v1.7.9 (see https://github.com/rhysd/actionlint/releases/tag/v1.7.9) using constants in `if` is now considered an error.

With this PR, I propose to ignore the error about using the constant `false` in an `if` for the file `.github/workflows/ci.yaml`.


7-zete-7:
CI / Tests failed due the #873.

dunglas:
Thanks

PR #876, opened 2025-12-04
Install xdebug through PIE

rvanlaak:
And take a leap in the future 🚀  It's a lot faster as through `install-php-extensions`.

XDebug 3.5 with PHP 8.5 support just was released, and mentions using PIE as a recommended method: https://github.com/xdebug/xdebug/releases/tag/3.5.0

Other extension I'm successfully installing for dev / production, by defining them in `composer.json` currently is `ext-zend-opcache`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,9 +22,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 	git \
 	&& rm -rf /var/lib/apt/lists/*
 
+COPY --from=ghcr.io/php/pie:bin /pie /usr/bin/pie
+COPY --from=composer /usr/bin/composer /usr/bin/composer
+
 RUN set -eux; \
 	install-php-extensions \
-		@composer \
 		apcu \
 		intl \
 		opcache \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -22,9 +22,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 	git \
 	&& rm -rf /var/lib/apt/lists/*
 
+COPY --from=ghcr.io/php/pie:bin /pie /usr/bin/pie
+COPY --from=composer /usr/bin/composer /usr/bin/composer
```
7-zete-7:
```suggestion
COPY --link --from=ghcr.io/php/pie:bin /pie /usr/bin/pie
COPY --link --from=composer /usr/bin/composer /usr/bin/composer
```

7-zete-7 (commented):
Looks like a good step forward for me!

7-zete-7:
Hi @rvanlaak!

Do you think it makes sense to add PIE to the FrankenPHP image? `install-php-extensions` is currently pre-installed there. I guess it would be convenient also to have PIE pre-installed.

dunglas:
Wouldn't it better to improve `install-php-extensions` to use Pie? See https://github.com/mlocati/docker-php-extension-installer/issues/1125

rvanlaak:
@dunglas good point, never thought of it like that. In short; yes!

Ref https://github.com/mlocati/docker-php-extension-installer/issues/1125

Issue #800, opened 2025-06-05
Running on WSL, I get  Error: Maximum execution time of 30 seconds exceeded


JesterHeads:
Hello everyone,

I'm running docker in WSL, and i'm trying to learn symfony, I use the image symfony-docker for a really small project.
I only have two controller with a few lines of code with 3 routes, those routes display a short text, but I get this error " Error: Maximum execution time of 30 seconds exceeded" most of the time and I don't understand why It takes so much time to only display one phrase

Thanks in advance for your answers

bkosun:
It seems you need to move your files to the WSL filesystem or use this trick to increase performance:
https://github.com/dunglas/symfony-docker/blob/main/compose.override.yaml#L11-L13

Related issue: https://github.com/dunglas/symfony-docker/issues/242, https://github.com/dunglas/symfony-docker/issues/505, https://github.com/dunglas/symfony-docker/issues/509, https://github.com/dunglas/symfony-docker/issues/553

bkosun:
It seems you need to move your files to the WSL filesystem or use this trick to increase performance:
https://github.com/dunglas/symfony-docker/blob/main/compose.override.yaml#L11-L13

Related issues: https://github.com/dunglas/symfony-docker/issues/242, https://github.com/dunglas/symfony-docker/issues/505, https://github.com/dunglas/symfony-docker/issues/509, https://github.com/dunglas/symfony-docker/issues/553

bkosun:
It seems you need to move your files to the WSL filesystem or use this trick to improve performance:
https://github.com/dunglas/symfony-docker/blob/main/compose.override.yaml#L11-L13

Related issues: https://github.com/dunglas/symfony-docker/issues/242, https://github.com/dunglas/symfony-docker/issues/505, https://github.com/dunglas/symfony-docker/issues/509, https://github.com/dunglas/symfony-docker/issues/553

JesterHeads:
Thanks a lot, moving my folder in the linux file system solved my problem !

Issue #609, opened 2024-04-19
Problem with "docker-entrypoint" under Windows


Moskito89:
When I try to start the project with `docker compose up --pull always -d --wait` as described, it runs well (but slow) under macOS. Under Windows, an error appears saying 

```
exec /usr/local/bin/docker-entrypoint: no such file or directory
```

I read some information about that file and checked the encoding and some other things, but all of them seem to be fine. Has anyone got that error too? Has the [`COPY`](https://github.com/dunglas/symfony-docker/blob/e022f7333dd09029c662e32a306a4b63b48a1d28/Dockerfile#L41) under Windows problems?

7-zete-7:
Hello @Moskito89!

Can you show content of your's `compose.yaml`, `compose.override.yaml` and `Dockerfile` files?

maxhelias:
For the slowness on MacOS, this has been improved on the latest versions of docker for mac. If that's not enough, have a look at these lines : https://github.com/dunglas/symfony-docker/blob/c7c788aca1080709613d1e3001ddabd7b7f4531d/compose.override.yaml#L13

For your Windows issue, I have no idea and no windows installation to try. The content of your file doesn't really help, it would be better to publish a reproducer

maxhelias:
It seems that there are already a number of answers that should help you on [Google](https://www.google.com/search?q=exec+%2Fusr%2Flocal%2Fbin%2Fdocker-entrypoint%3A+no+such+file+or+directory)

Moskito89:
Thank you for your advice, @maxhelias. I have already tried to remove the `vendor` folder, but could not notice any change.

I have already visited some of the sites you mentioned and, as described at the beginning, checked the issues discussed there. For example, the encoding of the file, line endings, the quotation marks, etc. None of this has helped — or rather, everything seems to be in order. I have not been able to find any errors.

I can try to set up a test repository tomorrow and will send it to you then. Thanks for your help in advance!

Moskito89:
I‘m sorry, @7-zete-7, but I can‘t. Docker tells me, that the container needs to be run, to execute a command. So it seems like it would fail even before. 

I tried the basic example on the same system, and it ran correctly. So the problem has something to do with my configuration. Maybe I‘ll find sone time do debug that step by step, but that’s currently not the case. 

Maybe we should close the issue but mark it as unresolved.

7-zete-7:
@Moskito89 try the command below to start container.

```shell
[...]

Issue #509, opened 2023-11-18
Symfony is very slow in dev mode


Saintchicky:
Hello everyone!
I am stuck when I uncomment the line - /app/vendor to improve Symfony's performance under Windows.
In the php container logs, I have this:
Fatal error: Uncaught LogicException: Symfony Runtime is missing. Try running "composer require symfony/runtime". in /app/bin/console:8
Indeed, we remove it from the bind mount, but then, I can't find the next step.
Thank you for your help!

dunglas:
Can you install the missing package?

Saintchicky:
Thank you for responding.
I do have my vendor installed with the runtime dependency.
If I don't uncomment the line - /app/vendor, the project works but it's very slow (the home page, which doesn't have any database queries, just a small jpg image: 19404ms, and when I refresh it's 9800ms by the Symfony Profiler)
An additional piece of information, I don't have composer and php installed on my computer.
I have installed Turbo, and when I do a scroll, it takes between 10000ms and 9000ms.
https://symfonycasts.com/screencast/symfony-doctrine/forever-scroll

ili101:
> Realistically what we would all want is that the `/app/vendor` folder is only synced one way: container -> host

@jurchiks I have a question because I work differently (with devcontainer). what is the point of having the vendor folder on the host and using Docker?
If you have already installed all the dependencies to run the project locally why do you also use Docker locally?

Also, I'm not familiar with Mac but on Windows, if you put your project folder in your WSL Ubuntu partition, mounting or not mounting the folder to Docker should have a similar performance.

jurchiks:
@ili101 I haven't tried putting the folder in WSL, guess I should try that. I read about it but somehow forgot...

As for devcontainer - I tried setting that up, but PHPStorm didn't (or maybe stll doesn't) support devcontainers, and I need the vendor folder for the code analyzer, otherwise everything's red.

ili101:
> @ili101 I haven't tried putting the folder in WSL, guess I should try that. I read about it but somehow forgot...
> 
> As for devcontainer - I tried setting that up, but PHPStorm didn't (or maybe stll doesn't) support devcontainers, and I need the vendor folder for the code analyzer, otherwise everything's red. The dumb thing still has no support for just reading the vendor folder contents from the container like VS Code does, afaik.

I made a fork with Devcontainer https://github.com/dunglas/symfony-docker/pull/523. works nicely with VSCode.
[...]

PR #879, opened 2025-12-17
Recommend using Linux under WSL for Symfony Docker on Windows

MindeedG:
On Windows, Docker Desktop can sometimes cause cache-related issues, which may lead to slower builds or unexpected behavior when working with Symfony Docker projects.

This PR updates the documentation to recommend using a Linux distribution under WSL (e.g., Ubuntu or Debian) and installing Docker natively in the Linux environment. This setup provides a more stable and reliable workflow.

It includes instructions for:

Installing Ubuntu under WSL
Installing Docker directly in Ubuntu

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -85,3 +85,24 @@ You can finally start your prod container(s) by running:
 ```console
 docker compose -f compose.yaml -f compose.prod.yaml up --wait
 ```
+
+## Cache issues on Windows
+
+On Windows, Docker Desktop can sometimes cause cache-related issues. To avoid this, it is recommended to use a Linux distribution under WSL (e.g., Ubuntu, Debian) when working with Docker and Symfony.
+
+Install Ubuntu on WSL
+
+```
+wsl --install -d Ubuntu
+```
+
+Install Docker on Ubuntu
+
+```
+sudo apt update
+sudo apt install -y docker.io
+sudo systemctl enable --now docker
+sudo usermod -aG docker $USER
+```
+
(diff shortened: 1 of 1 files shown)


7-zete-7:
Hi @MindeedG!

Wow! I haven't seen any information about cache-related issues when using Docker Desktop on WSL2 before. Do you have any links to resources that describe this issue? I'd be interested in learning more about it.

MindeedG:
Hi @7-zete-7, 

I don't have any official resources to share with you, but after encountering these issues, AI suggested this solution to me, and it worked.

It seems to be related to the fact that this installation doesn't share filesystem between OS.

7-zete-7:
I tried to find any information about this issue, but I couldn't find any mention of it.

It currently looks like [the XY problem](https://xyproblem.info/). Cache and performance issues with Windows/WSL are occasionally encountered, but they're usually related to an incorrectly configured environment or configuration. It could also be because the project was located on the Windows filesystem, not in WSL2 itself (this does cause problems, see #800, #609, #509, etc.).

If there really is a cache-related issue with Docker Desktop, I think it would be best to report it in the [Docker Desktop for Windows repository](https://github.com/docker/for-win/issues) so that the issue can be brought to light.

tcoch:
I agree on the principle : I have a Windows worstation, and without WSL2, it's a pain in the A.
But this applies to every project / container you will run on Docker, therefore, it's a `Docker `problem, not a `Symfony Docker` problem
IMHO This information is too generic to be part of the documentation

PR #880, opened 2025-12-17
fix: Add `var/share` directory creation

7-zete-7:
Since Symfony 7.4, a new `var/share` directory has been introduced for storing files shared between servers (see https://symfony.com/blog/new-in-symfony-7-4-share-directory).

This PR adds pre-creation of this new directory, similar to the `var/cache` and `var/log` directories.


dunglas:
Thanks!

PR #890, opened 2026-01-27
bug: set max_execution_timeout to -1 for Xdebug

gimler:
Disable timeout for Xdebug to allow breakpoints.

see also https://github.com/dunglas/symfony-docker/discussions/889


maxhelias:
It is not good practice to disable this option. Therefore, it is not possible to integrate it here in this template.

maxhelias:
Just to be clear, this does not prevent you from using it temporarily for xdebug. But this cannot be a general rule. Thank you.

PR #891, opened 2026-01-29
fix: Use built-in Mercure

7-zete-7:
After https://github.com/symfony/recipes/pull/1514 is accepted, it will be possible to use FrankenPHP's built-in Mercure without setting the `MERCURE_URL` environment variable.

Also eliminating the need for a _loopback_ configuration to access Mercure also eliminates the need for the `php:80` address.


7-zete-7:
Closed due #869 duplication.

PR #893, opened 2026-01-29
fix(lint): Ignore always-false condition errors in ci workflow again

7-zete-7:
Actionlint's [previously ignored](https://github.com/dunglas/symfony-docker/pull/874) spurious error message has been changed (see https://github.com/rhysd/actionlint/commit/524fc0af10c1c20fec96b53075b537a1df694890).


maxhelias:
I'm fixing it in a different way here : https://github.com/dunglas/symfony-docker/pull/895

PR #911, opened 2026-03-12
fix: Make sure `docker-entrypoint.sh` being cleaned

7-zete-7:
I noticed some problems with the [original idea](https://github.com/dunglas/symfony-docker/pull/908):

- `docker-entrypoint.sh` only changes within the container.
- Entire block wasn't deleted.
- `sed` couldn't replace the original file with the modified one.

In this PR, I propose:

- Clarify the `sed` expression.
- Run `sed` on the source file `frankenphp/docker-entrypoint.sh`
- Added removal of the `SYMFONY_VERSION` and `STABILITY` envs from `compose.yaml`
- Switch to Flex-like markers

According to my tests, these changes lead to the expected behavior.

The only thing that remains unclear is that the first dev image will contain the original `docker-entrypoint.sh` content. However, this shouldn't affect the container's behavior.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -15,7 +15,7 @@ services:
       MERCURE_PUBLIC_URL: ${CADDY_MERCURE_PUBLIC_URL:-https://${SERVER_NAME:-localhost}:${HTTPS_PORT:-443}/.well-known/mercure}
       MERCURE_JWT_SECRET: ${CADDY_MERCURE_JWT_SECRET:-!ChangeThisMercureHubJWTSecretKey!}
       ###> dunglas/symfony-docker ###
-      # The two next lines used for initial installation
+      # Next lines are only used during initial installation
       # This block will remove itself after the installation
       SYMFONY_VERSION: ${SYMFONY_VERSION:-}
       STABILITY: ${STABILITY:-stable}


7-zete-7:
IMO the `SYMFONY_VERSION` and `STABILITY` envs should also be removed from `compose.yaml` file. But I can't see how to do this from within the container without mounting the `compose.yaml` file into the container, which doesn't seem like a good idea.

dunglas:
Good catch. But then we should also remove the mountpoint. This starts to be complicated. Maybe I should just revert #908. What do you think?

7-zete-7:
@dunglas I agree, the mounting will need to be removed too.

I don't think it need to be reverted. I guess it might work.

I just realized that there's no need to mount the `docker-entrypoint.sh` and `compose.yaml` files since they're already mounted in the container with the project directory in the dev env. I'll update the PR.

dunglas:
Nice! Btw, maybe we could switch to Flex-like markers for consistency?

7-zete-7:
@dunglas I like this idea.

Like `###> symfony-docker/project-install ###` and `###< symfony-docker/project-install ###`. WDYT?

7-zete-7:
Something else is needed. The `symfony-docker` org [already exists on Packagist](https://packagist.org/packages/symfony-docker/). It might cause problems in the future.

`###> dunglas/symfony-docker-project-install ###` and `###< dunglas/symfony-docker-project-install ###`?

dunglas:
or just `###> dunglas/symfony-docker ###` so we're sure there is no conflict?

dunglas:
Thanks!

PR #932, opened 2026-04-20
feat: add file matcher on assets and precompress response

94noni:
related to the `php bin/console asset-map:compile`
and doc https://symfony.com/doc/current/frontend/asset_mapper.html#pre-compressing-assets

wdyt? is it worth it?

edit: I am just unsure of the `root` and `path` logic :s

### DIFF
--- a/frankenphp/Caddyfile
+++ b/frankenphp/Caddyfile
@@ -44,6 +44,11 @@
 	# Disable Topics tracking if not enabled explicitly: https://github.com/jkarlin/topics
 	header ?Permissions-Policy "browsing-topics=()"
 
+	@assets path /assets/*
+	file_server @assets {
+		precompressed zstd br gzip
+	}
+
 	@phpRoute {
 		not path /.well-known/mercure*
 		not file {path}


dunglas:
Couldn't you just add `precompressed` to the existing `file_server` directive?

dunglas:
That would also be nice to add some docs about this new feature.

dunglas:
Actually, this will add a performance penalty for everybody not using this feature. I would just add a comment pointing to the docs. WDYT?

94noni:
> Actually, this will add a performance penalty for everybody not using this feature. I would just add a comment pointing to the docs. WDYT?

make sens, as in the dockerfile its also a condition for the "`asset-map:compile`" part

can you just confirm me that the pattern of the match is ok ?
with the above `root /app/public` that we dont have to write `@assets path /app/public/assets/*`

i will revisit this PR soon !

94noni:
closing for now, will reopen a PR with doc link here and of course opening a frankenphp repo doc PR

94noni:
the doc exists at https://frankenphp.dev/docs/symfony/#pre-compressing-assets
i will just add a comment on the file

PR #937, opened 2026-04-25
refactor(php): remove unnecessary `opcache.preload_user` overwrite

nextpageblog:
Closes https://github.com/dunglas/symfony-docker/discussions/936.


dunglas (approved):

dunglas:
Good catch!

PR #938, opened 2026-04-25
fix(docs): clarify where to apply Symfony pre-7.4 worker mode setup steps

### LINKED ISSUE #934: Symfony 6.x incompatible with FrankenPHP worker mode since commit `1ad09f7`
Hi!

Since [commit 1ad09f7: "Remove unneeded `runtime/frankenphp-symfony` dependency"](https://github.com/dunglas/symfony-docker/commit/1ad09f7), generating a new project with `SYMFONY_VERSION=6.*` results in a broken setup where FrankenPHP never starts (`FrankenPHP started 🐘` never appears in logs) and the application is unreachable.

## Steps to reproduce

```bash
SYMFONY_VERSION=6.* docker compose up
```

## Expected behavior

**FrankenPHP** starts correctly and the application is accessible at `https://localhost`, as it does with **Symfony 7 and 8**.

### ❌ Symfony 6 (broken)

**FrankenPHP** fails to start (`FrankenPHP started 🐘` never appears in logs) and the application is unreachable on `port 443`.

The following errors are repeatedly logged in the terminal, interleaved with the HTML content of the error page dumped multiple times:

```
INFO    http.handlers.php    [error] Uncaught PHP Exception Symfony\Component\HttpKernel\Exception\NotFoundHttpException: "No route found for "GET http:///index.php/"" at RouterListener.php line 127
```

When running the following command directly inside the container:

```bash
docker compose run --rm php php public/index.php 2>&1
```

The actual root cause becomes visible:

```
Uncaught PHP Exception Symfony\Component\HttpKernel\Exception\BadRequestHttpException: "Invalid Host "localhost, php"." at HttpKernel.php line 83
```

### ✅ Symfony 7 and 8 (working)

**FrankenPHP** starts correctly and the following appears in the logs:

```
INFO    frankenphp      FrankenPHP started 🐘   {"php_version": "8.5.5", "num_threads": 49, "max_threads": 49}
INFO    http    enabling HTTP/3 listener        {"addr": ":443"}
INFO    http.log        server running  {"name": "srv0", "protocols": ["h1", "h2", "h3"]}
INFO    http.log        server running  {"name": "srv1", "protocols": ["h1", "h2", "h3"]}
```

## Root cause

Commit `1ad09f7` removed two things simultaneously:
1. `runtime/frankenphp-symfony` from the `composer require` command in `docker-entrypoint.sh`
2. `env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime` from the Caddyfile worker config

**Symfony 7** automatically detects **FrankenPHP** and uses the correct runtime without these. **Symfony 6** does not have this capability and requires both to work correctly in worker mode.

## Investigated approaches

The following approaches were investigated but did not fully resolve the issue:

### 1. Installing `runtime/frankenphp-symfony` separately

Adding `composer require runtime/frankenphp-symfony` as a separate step after the initial `composer require "php:>=$PHP_VERSION"` in `docker-entrypoint.sh` installs the package correctly, but without restoring `env APP_RUNTIME` in the Caddyfile, **FrankenPHP** still fails to start.

### 2. Fixing `config/packages/routing.yaml`

The **Symfony 7 Flex** recipe for `symfony/routing` (`>=7.4`) generates a `routing.yaml` with `default_uri` configured, while the **Symfony 6** recipe (`>=6.2`) leaves it commented out. Uncommenting and setting `default_uri: '%env(DEFAULT_URI)%'` in `routing.yaml` and adding `DEFAULT_URI=http://localhost` to `.env` did not resolve the issue on its own.

### 3. Configuring `trusted_hosts` in `framework.yaml`

Adding `trusted_hosts` to `config/packages/framework.yaml` to accept `localhost` and `php` as valid hosts did not resolve the issue on its own.

### 4. Separating `SERVER_NAME` from the internal Caddy vhost

The `SERVER_NAME` variable (`localhost, php:80`) serves a dual purpose: defining **Caddy** vhosts and being used internally by **FrankenPHP**. Separating these concerns via a dedicated variable in the Caddyfile was considered but would require a more significant refactor of the Caddyfile structure.

## Workaround

A working setup for **Symfony 6** is available on the [`symfony6-frankenphp-compatibility`](https://github.com/jprivet-dev/symfony-docker/tree/symfony6-frankenphp-compatibility) branch of my fork. You can use it directly to generate a **Symfony 6** project without issues:

```bash
git clone https://github.com/jprivet-dev/symfony-docker.git symfony-docker-jpr -b symfony6-frankenphp-compatibility
cd symfony-docker-jpr
SYMFONY_VERSION=6.* docker compose up
```

> **NOTE :** The `symfony-docker-jpr` folder name is used to avoid conflicts with an existing `symfony-docker` clone from the original repository.

The two changes compared to `main` are:

**`frankenphp/docker-entrypoint.sh`** — restore `runtime/frankenphp-symfony` in the `composer require` call:
```sh
composer require "php:>=$PHP_VERSION" runtime/frankenphp-symfony
```

**`frankenphp/Caddyfile`** — restore `env APP_RUNTIME` in the worker block:
```caddyfile
worker {
    file ./public/index.php
    env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime
    {$FRANKENPHP_WORKER_CONFIG}
}
```

✅ With the fix applied, **Symfony 6** also starts correctly:

```
INFO    frankenphp      FrankenPHP started 🐘   {"php_version": "8.5.5", "num_threads": 49, "max_threads": 49}
INFO    http    enabling HTTP/3 listener        {"addr": ":443"}
INFO    http.log        server running  {"name": "srv0", "protocols": ["h1", "h2", "h3"]}
INFO    http.log        server running  {"name": "srv1", "protocols": ["h1", "h2", "h3"]}
```

## Suggested fix

Honestly, I’ve no idea! 😅 I’ve tried a few different methods, but without success (see above). What do you suggest?

jprivet-dev:
Closes #934

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -31,11 +31,14 @@ set SYMFONY_VERSION=6.4.* && docker compose up --wait&set SYMFONY_VERSION=
 >
 > If you're using Symfony 7.3 or earlier with FrankenPHP in worker mode, you also need to follow these steps
 >
-> ```console
-> composer require runtime/frankenphp-symfony
+> In `frankenphp/docker-entrypoint.sh`, add `runtime/frankenphp-symfony` right after the existing `composer require` line:
+>
+> ```diff
+>   composer require "php:>=$PHP_VERSION"
+> + composer require runtime/frankenphp-symfony
 > ```
 >
-> Add `env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime` in the `frankenphp/Caddyfile` in the `worker` section.
+> In `frankenphp/Caddyfile`, add `env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime` in the `worker` section:
 >
 > ```diff
 >  worker {
--- a/docs/options.md
+++ b/docs/options.md
@@ -11,7 +11,47 @@ You can customize the Docker build process using these environment variables.
 
 Use the `SYMFONY_VERSION` environment variable to select a specific Symfony version.
 
-For instance, use the following command to install Symfony 6.4:
+### Install Symfony 7.4+
+
+On Linux:
+
+```console
+SYMFONY_VERSION=7.4.* docker compose up --wait
+```
+
+On Windows:
+
+```console
+set SYMFONY_VERSION=7.4.* && docker compose up --wait&set SYMFONY_VERSION=
+```
+
+### Install Symfony pre-7.4
+
+If you're using Symfony 7.3 or earlier with FrankenPHP in worker mode,
+you need to prepare two files **before** running `docker compose up`.
+
+In `frankenphp/docker-entrypoint.sh`, add `runtime/frankenphp-symfony` to the existing `composer require` line:
+
+```diff
+- composer require "php:>=$PHP_VERSION"
++ composer require "php:>=$PHP_VERSION" runtime/frankenphp-symfony
+```
+
+In `frankenphp/Caddyfile`, add `env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime` in the `worker` section:
+
+<!-- markdownlint-disable MD010 -->
+
+```diff
+ worker {
+ 	file ./public/index.php
++	env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime
+ 	{$FRANKENPHP_WORKER_CONFIG}
+ }
+```
+
+<!-- markdownlint-enable MD010 -->
+
+Then start the project:
 
 On Linux:
 
@@ -24,31 +64,6 @@ On Windows:
 ```console
 set SYMFONY_VERSION=6.4.* && docker compose up --wait&set SYMFONY_VERSION=
 ```
-
-<!-- markdownlint-disable MD010 -->
-
-> [!NOTE]
->
-> If you're using Symfony 7.3 or earlier with FrankenPHP in worker mode, you also need to follow these steps
->
-> In `frankenphp/docker-entrypoint.sh`, add `runtime/frankenphp-symfony` right after the existing `composer require` line:
->
-> ```diff
->   composer require "php:>=$PHP_VERSION"
-> + composer require runtime/frankenphp-symfony
-> ```
->
-> In `frankenphp/Caddyfile`, add `env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime` in the `worker` section:
->
-> ```diff
->  worker {
->  	file ./public/index.php
-> +	env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime
->  	{$FRANKENPHP_WORKER_CONFIG}
->  }
-> ```
-
-<!-- markdownlint-enable MD010 -->
 
 ## Installing Development Versions of Symfony
 


jprivet-dev:
By the way, I went with two separate `composer require` calls for clarity:

```diff
  composer require "php:>=$PHP_VERSION"
+ composer require runtime/frankenphp-symfony
```

But I'm happy to switch to a single line if you prefer:

```diff
- composer require "php:>=$PHP_VERSION"
+ composer require "php:>=$PHP_VALUE" runtime/frankenphp-symfony
```

Let me know! 🙂

bkosun (commented):
Please note that the section of the `frankenphp/docker-entrypoint.sh` file that you suggest modifying is used to initialize the project and will be automatically removed after the first run:

https://github.com/dunglas/symfony-docker/blob/777be58d05333c40642a7d530b45ae5b376d7dc7/frankenphp/docker-entrypoint.sh#L20-L22

Support for Symfony 7.0–7.3 has ended, and support for Symfony 6.4 will end in November 2026. Do you still want to start a new project using one of these versions?

https://symfony.com/releases

If you want to integrate this template into an existing project, you first need to manually install the `runtime/frankenphp-symfony` package and modify the configuration as described in the documentation:

> The worker mode of FrankenPHP is enabled by default. To use it with Symfony ≤ 7.3, install the FrankenPHP runtime:
> 
> `composer require runtime/frankenphp-symfony`
> 
> Then update worker configuration:
> 
> ```diff
>  worker {
>  	file ./public/index.php
> +	env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime
>  	{$FRANKENPHP_WORKER_CONFIG}
>  }
> ```

https://github.com/dunglas/symfony-docker/blob/main/docs/existing-project.md

By the way, some users may prefer the classic mode, so the `runtime/frankenphp-symfony` package is optional.

bkosun (commented):
Please note that the section of the `frankenphp/docker-entrypoint.sh` file that you suggest modifying is used to initialize the project and will be automatically removed after the first run:

https://github.com/dunglas/symfony-docker/blob/777be58d05333c40642a7d530b45ae5b376d7dc7/frankenphp/docker-entrypoint.sh#L20-L22

Support for Symfony 7.0–7.3 has ended, and support for Symfony 6.4 will end in November 2026. Do you still want to start a new project using one of these versions?

https://symfony.com/releases

If you want to integrate this template into an existing project, you first should to manually install the `runtime/frankenphp-symfony` package and modify the configuration as described in the documentation:

> The worker mode of FrankenPHP is enabled by default. To use it with Symfony ≤ 7.3, install the FrankenPHP runtime:
> 
> `composer require runtime/frankenphp-symfony`
> 
> Then update worker configuration:
> 
> ```diff
>  worker {
>  	file ./public/index.php
> +	env APP_RUNTIME Runtime\FrankenPhpSymfony\Runtime
>  	{$FRANKENPHP_WORKER_CONFIG}
>  }
> ```

https://github.com/dunglas/symfony-docker/blob/main/docs/existing-project.md

By the way, some users may prefer the classic mode, so the `runtime/frankenphp-symfony` package is optional.

jprivet-dev:
Hi @bkosun!

> Support for Symfony 7.0–7.3 has ended, and support for Symfony 6.4 will end in November 2026. Do you still want to start a new project using one of these versions?

That's a fair question! I wouldn't start a long-lived production project with **Symfony 6.4** either.

That said, my use case is different: quickly spinning up a **reproducer** for a bug report or a compatibility issue. I needed to do exactly that twice last week — once for `SMimeSigner` and once for `GotenbergBundle`. In both cases, `SYMFONY_VERSION=6.4.*` is the most straightforward way to reproduce the issue in isolation.

**Symfony Docker** is a great tool for this kind of workflow, and **Symfony 6.4** is still officially maintained until **November 2026** — so these reproduction environments represent a practical and effective use case.

What makes this PR low-risk is that the fix is only **2 lines** across two files — no new compatibility layer to maintain, just documentation that reflects what actually works. I've restructured the section into `Symfony 7.4+` and `Symfony pre-7.4` subsections to make the required steps and their order explicit before running `docker compose up`.

Ultimately, this is a matter of philosophy: do you want to keep `SYMFONY_VERSION` as a general-purpose option that works for any maintained version, or do you prefer to restrict it to **Symfony 7.4+** and drop any reference to older versions entirely?

I just want to make sure the doc matches whichever decision you make. 🙂

jprivet-dev:
Hi @bkosun!

What choice do we make? :)

7-zete-7 (approved):

maxhelias (approved):

maxhelias:
Thanks @jprivet-dev !

PR #949, opened 2026-06-19
feat(devcontainer): default to Kilo Code, make Claude Code opt-in

### LINKED ISSUE #943: Devcontainer introduces opinionated AI tooling and agent setup in a project template
First of all, thank you for maintaining this project. Symfony Docker is widely used as a starting point for new Symfony applications and provides a very convenient baseline environment.

While reviewing the repository I noticed that the `.devcontainer` configuration introduces several components related to Claude Code and AI agent workflows.

The devcontainer installs Claude Code via a devcontainer feature and also includes additional infrastructure around it. For example, the setup configures a firewall that restricts outbound network access from the container.

In addition, when the project is opened in the devcontainer, initialization scripts create symlinks in the workspace pointing to `.claude` and introduce files such as `CLAUDE.md` and `AGENTS.md`.

Taken together, the devcontainer does more than describe a development environment and effectively prepares the workspace for a specific AI-assisted development workflow.

This feels somewhat surprising in the context of a project template. Templates like Symfony Docker are often used as a neutral starting point for many teams and development setups, so it is generally helpful if they focus primarily on providing the runtime environment (Docker, PHP, Caddy, etc.) without prescribing specific developer tooling or workflows.

Including Claude Code integration, agent manifests, firewall sandboxing, and workspace symlinks implicitly promotes a particular tool and workflow. While these features may be useful for some users, they are not directly related to running a Symfony application and may be unexpected for teams that simply want a Docker-based Symfony environment.

It might be worth considering making this integration optional instead of part of the default template. For example, the AI-related setup could live in an alternative devcontainer configuration or be documented as an optional addition for teams interested in AI-assisted workflows.

This would keep the main template focused on Symfony and Docker while still allowing users who want to experiment with Claude Code or agent-based workflows to opt in easily.

Thanks again for the work on the project.

dunglas:
Closes #943 

Ship the open source Kilo Code extension as the pre-installed agent. It drives local models (Ollama/LM Studio via the host gateway) or remote providers, so the sandbox no longer needs Anthropic's domains by default. Document how to add Claude Code back in docs/agents.md.

### DIFF
--- a/.devcontainer/.claude/settings.json
+++ b/.devcontainer/.claude/settings.json
@@ -1,5 +1,0 @@
-{
-    "enabledPlugins": {
-        "php-lsp@claude-plugins-official": true
-    }
-}
--- a/.devcontainer/AGENTS.md
+++ b/.devcontainer/AGENTS.md
@@ -13,7 +13,7 @@ If an outbound request fails (e.g., `curl`, `composer require`, `npm install` to
 Edit `.devcontainer/init-firewall.sh` and add the domain to the `ipset=` line in the dnsmasq configuration block:
 
 ```bash
-ipset=/github.com/anthropic.com/.../NEW_DOMAIN.COM/allowed-domains
+ipset=/github.com/.../NEW_DOMAIN.COM/allowed-domains
 ```
 
 Then rebuild the Dev Container to apply the change.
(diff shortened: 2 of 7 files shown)


dunglas:
Superseded by https://github.com/dunglas/symfony-docker/pull/950, which removes preinstalled coding-agent support entirely and documents an opt-in setup instead of defaulting to Kilo Code.

PR #954, opened 2026-06-30
Add PDO driver for PostgreSQL

mbobby22:
Trying to setup the default PostgreSQL database as in the documentation here:

https://symfony.com/doc/current/the-fast-track/en/7-database.html

Results in PHP container log errors:

```
Still waiting for database to be ready... Or maybe the database is not reachable. 59 attempts left.
```

The PHP container was failing because the PostgreSQL PDO extension was missing inside the image and Symfony’s readiness check reported: `could not find driver`

I rebuilt the PHP image and verified the fix and the container now has pdo_pgsql installed and loaded
the Symfony readiness check completed successfully with: `The database is now ready and reachable`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,9 @@ ENV COMPOSER_ALLOW_SUPERUSER=1
 ENV PHP_INI_SCAN_DIR=":$PHP_INI_DIR/app.conf.d"
 
 ###> recipes ###
+###> doctrine/doctrine-bundle ###
+RUN install-php-extensions pdo_pgsql
+###< doctrine/doctrine-bundle ###
 ###< recipes ###
 
 COPY --link frankenphp/conf.d/10-app.ini $PHP_INI_DIR/app.conf.d/


7-zete-7:
Hi, @mbobby22!

This isn't necessary, as the `pdo_pgsql` extension is installed automatically via the `doctrine/doctrine-bundle` recipe (see https://github.com/symfony/recipes/blob/8b4def21b69d6d8351626b62598ca84962ce226f/doctrine/doctrine-bundle/3.0/manifest.json#L18-L20).

Important note: After installing the `doctrine/doctrine-bundle` package (and, consequently, the corresponding recipe), you must rebuild the container with the `docker compose build php` command.

As I see, the article https://symfony.com/doc/current/the-fast-track/en/7-database.html omitted information about adding the extension and rebuilding the image.

mbobby22:
Hi,

What I did was use this command:

```
composer require orm
# Symfony Flex recipe alias that installs:
# - doctrine/doctrine-bundle
# - doctrine/orm
# - doctrine/dbal
```
It includes de `doctrine/doctrine-bundle` recipe, but for some reason the driver did not get installed.
I will do some more testing and confirm if the issue persists.

Thank you for the review so far!

PR #958, opened 2026-07-15
Refactor migration check condition in entrypoint script

axzx:


frankenphp/docker-entrypoint.sh
```diff
@@ -58,7 +58,7 @@ if [ "$1" = 'frankenphp' ] || [ "$1" = 'php' ] || [ "$1" = 'bin/console' ]; then
 			echo 'The database is now ready and reachable'
 		fi
 
-		if [ "$(find ./migrations -iname '*.php' -print -quit)" ]; then
+		if find ./migrations -iname '*.php' -print -quit | grep -q .; then
```
7-zete-7: It's better to use full option names to make it obvious.
```suggestion
		if find ./migrations -iname '*.php' -print -quit | grep --quiet .; then
```

7-zete-7 (commented):
Hi @axzx!

I agree, the current condition doesn't seem obvious.

7-zete-7 (approved):

zpi12lmm:
I think since changes are being made to migrations, it's also worth adding a condition that migrations are only performed for the main PHP service (usually create from the main PHP service other worker services that have the same entry point, and the execution of migrations can compete with each other).

maxhelias (approved):

maxhelias:
Thanks @zpi12lmm !

axzx:
> Thanks @zpi12lmm !

@maxhelias thanks for merge:)

maxhelias:
Oh, sorry, I confused the usernames. Thanks @axzx 😄

PR #961, opened 2026-08-27
chore: disable black and ruff on CI

alexisLefebvre:
It reported a warning:

> [WARN]   Black and Ruff are both enabled, and might conflict with each other. To avoid potential conflicts, keep only one of the two enabled, and disable the other.

We do not need to lint Python files.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -82,3 +82,5 @@ jobs:
           VALIDATE_TRIVY: false
           VALIDATE_BIOME_FORMAT: false
           VALIDATE_BIOME_LINT: false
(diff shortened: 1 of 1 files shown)


maxhelias (approved):

maxhelias:
Thanks @alexisLefebvre !

dunglas:
I would enable one of them. Python scripts are very common and the whole point of having super linter in this repo is to lint all common files a project can have without having to tweak the config or install extra tools.

maxhelias:
What do you think about keeping Ruff?

dunglas:
Yes good idea

alexisLefebvre:
Ok, Ruff will be restored by this PR:

- #963

PR #962, opened 2026-08-27
chore: reduce verbosity of super-linter

alexisLefebvre:
The default output of super-linter is too verbose.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -78,6 +78,9 @@ jobs:
         env:
           GITHUB_ACTIONS_CONFIG_FILE: actionlint.yaml
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


alexisLefebvre:
Mhh, the long output is a side-effect of using this template on GitLab CI, where sections are not collapsed, while they are on GitHub: https://github.com/dunglas/symfony-docker/actions/runs/33101510794/job/98620144522?pr=961

It reduced the output on GitHub though: https://github.com/dunglas/symfony-docker/actions/runs/33102275562/job/98622828668?pr=962

maxhelias:
I'm not convinced by these changes if they're just to deal with a side effect of the GitLab CI :thinking:

alexisLefebvre:
Fair enough, for some reason it is in fact worst:

Before:

> image

After:

> image

Let’s keep it as-is.

PR #948, opened 2026-06-12
docs: add documentation page about sending and testing emails

enginydigital:
## What

Adds a new `docs/mailer.md` page covering how email works with this template:

- How the Symfony Flex recipe wires Mailpit in `compose.override.yaml` when
  installing `symfony/mailer`, and how to access its web UI (the ports are
  exposed without a fixed host mapping, which regularly confuses users).
- The correct `MAILER_DSN` for development (`smtp://mailer:1025`), with a
  warning about the common `127.0.0.1` pitfall (#723).
- Production setup: the Mailpit service only exists in
  `compose.override.yaml`, so nothing intercepts emails in production and a
  real transport must be configured via `MAILER_DSN` (#830, discussion #831).
- A short note about reusing Mailpit on staging environments as a safety net.

## Why

Email configuration is a recurring source of confusion in the issue tracker
(#225, #723, #830, discussion #831). The Flex recipe does the heavy lifting,
but none of it is documented here, so users don't know Mailpit is already
available, how to reach it, or what changes in production.

The page follows the style of the existing docs (`mysql.md`, `xdebug.md`) and
is linked from the README docs index (added after "Support for extra services"):

```markdown
4. [Sending and testing emails with Mailpit](docs/mailer.md)
```

PR #966, opened 2026-09-14
docs: document how to use another email catcher

tsokolovs:
## What

Adds a "Using Another Email Catcher" section to `docs/extra-services.md`.
It shows how to point the recipe-installed `mailer` service at a different SMTP
catcher, with [Mailtrap Local](https://github.com/mailtrap/mailtrap-local) as
the example.

Documentation only. No change to any compose file, and no new page.

## Why

`docs/extra-services.md` records that `symfony/mailer` installs a Mailpit
service. The page does not say that the service is replaceable, or what a
replacement needs. The recipe writes the service into `compose.override.yaml`,
so the swap is a two-key edit plus the DSN.

The section also answers the port question that applies to any catcher in this
template. The recipe exposes the ports without a fixed host mapping, so
`docker compose port` is the way to reach the web interface.

## Verified

Docker Compose v5.3.0 with `mailtrap/mailtrap-local` 0.4.1:

- `docker compose up --wait` reports the container healthy.
- A message sent from a second container to `mailer:3535` arrives.
- `GET /api/v1/messages` on port 3550 lists that message.
- `GET /` on port 3550 returns the web interface.
- `docker compose port mailer 3550` returns the host address.

The image binds `0.0.0.0` through its own default command, so the block needs no
`command:` key. The tool reads no environment variables, so the block needs no
`environment:` key either.

`super-linter/slim@v8` ran locally with the same disabled linters as `ci.yaml`.
It reports "All files and directories linted successfully".


tsokolovs:
For context: #948 adds a dedicated `docs/mailer.md` page for the same topic. It is approved but not merged yet, so this PR targets `docs/extra-services.md`, which exists today.

If #948 merges first, `docs/mailer.md` is the better home for this content. I have the equivalent section prepared against that page as a 25-line follow-up. I am happy to move it there and reduce this section to a one-line pointer, or to close this PR in favor of that one. Whichever you prefer.

PR #967, opened 2026-09-22
fix: npm warn install-scripts

fastnloud:
This should fix the following warning:

```bash
npm warn install-scripts 1 package has install scripts not yet covered by allowScripts:
npm warn install-scripts   protobufjs@8.3.0 (postinstall: node scripts/postinstall)
npm warn install-scripts
npm warn install-scripts Run `npm install -g --allow-scripts=protobufjs` to allow these scripts once, or `npm config set allow-scripts=protobufjs --location=user` to allow them for all global installs.
```


7-zete-7:
Hi @fastnloud!

It looks like this issue has already been reported in https://github.com/protobufjs/protobuf.js/issues/2296 and fixed in https://github.com/protobufjs/protobuf.js/pull/2299. The `postinstall` script no longer exists since `protobufjs` v8.6.0. `intellephense` uses `protobufjs` v8.3.0, which had the same functionality of the the `postinstall` script. I'm guessing adding this option isn't necessary.

fastnloud:
Thanks @7-zete-7 for looking into it. I think you might be right and that it'll fix itself whenever `intellephense` decides to up this dependency. 

Feel free to close this if no longer relevant.

maxhelias:
Thank you both

